PR #11, opened 2020-08-11
Create ComputeAsset class

svenvanderburg:
Create ComputeAsset class, including changes to all code related to compute assets

### DIFF
--- a/proof_of_concept/asset.py
+++ b/proof_of_concept/asset.py
@@ -1,0 +1,51 @@
+"""Classes for describing assets."""
+from typing import Any, Dict
+
+
+class Asset:
+    """Asset, a representation of a computation or piece of data."""
+
+    def __init__(self, name: str, data: Any, metadata: Any):
+        """Constructor.
+
+        TODO: typehint Metadata class for metadata argument, problematic
+            due to circular imports
+        Args:
+            name: Name of the asset
+            data: Data related to the asset
+            metadata: Metadata related to the asset
+        """
+        self.name = name
+        self.data = data
+        self.metadata = metadata
+
+
+class ComputeAsset(Asset):
+    """Compute asset, represents a computing step, i.e. software."""
+
+    def run(self, inputs: Dict[str, Any]) -> Dict[str, Any]:
+        """Run compute step.
+
+         For now this executes one of several simple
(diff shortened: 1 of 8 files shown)


proof_of_concept/asset.py
```diff
@@ -0,0 +1,51 @@
...
+from typing import Any, Dict
+
+
+class Asset:
+    """Asset, a representation of a computation or piece of data."""
+
+    def __init__(self, name: str, data: Any, metadata: Any):
+        """Constructor.
+
+        TODO: typehint Metadata class for metadata argument, problematic
```
svenvanderburg, line 11: Fixing this would require quite some refactoring, would maybe be nice to do together, or ignore it for now.

proof_of_concept/asset.py
```diff
@@ -0,0 +1,51 @@
...
+            due to circular imports
+        Args:
+            name: Name of the asset
+            data: Data related to the asset
+            metadata: Metadata related to the asset
+        """
+        self.name = name
```
svenvanderburg, line 18: I am curious whether these attributes are what you have in mind @LourensVeen. Data and metadata don't seem to be really used yet.

svenvanderburg:
There are 2 failing tests, but they also fail in the 'develop' branch. Shall I fix them in a separate ticket/branch?

proof_of_concept/asset.py
LourensVeen: Come to think of it, maybe an IMetadata interface would be the simplest fix? I don't like interfaces with only one implementation, but there are a few other instances of this here already anyway, and for some reason this whole thing really wants to be Javaesque anyway, so...

proof_of_concept/asset.py
LourensVeen: The data would eventually be a reference to a container image containing the data, I think it now just contains an arbitrary Python object.

I think there's Metadata being used in the AssetStore. (Intermediate) results are `DataAsset`s, and they need Metadata for provenance and also for security, as what you can do with that intermediate result depends on how it was calculated.

LourensVeen:
Okay, looks great, and passes all the tests. Nice work! Going to merge it now.

PR #12, opened 2020-08-11
Fix mypy failure by ingoring line

svenvanderburg:


proof_of_concept/policy_replication.py
```diff
@@ -84,7 +84,7 @@ def _update(self) -> None:
         # removed ones that disappeared
         removed_servers = [
                 server for server in self._other_stores
-                if server not in list(zip(*new_servers))[1]]
+                if server not in list(zip(*new_servers))[1]]  # type: ignore
```
svenvanderburg, line 87: Mypy is complaining here with: "error: Unsupported right operand type for in ("object")". From a small inspection this code does not seem wrong.

svenvanderburg: i.e. the result of `list(zip(*new_servers))[1]` is a tuple

LourensVeen: Agreed, this looks like a mypy issue, it has these things from time to time.

LourensVeen (approved):
Looks fine, lgtm.

PR #14, opened 2020-08-19
Retrieve compute assets from assetstore instead of passing them to the workflow

svenvanderburg:


proof_of_concept/local_workflow_runner.py
```diff
@@ -174,6 +179,13 @@ def _get_step_inputs(
 
         return step_input_data
 
+    def _retrieve_compute_asset(self, compute_asset_name: str) -> ComputeAsset:
+        # TODO: the store id is hardcoded so scenario pii test works,
```
svenvanderburg: @LourensVeen would be nice to discuss this

tests/test_scenarios.py
```diff
@@ -105,31 +105,32 @@ def test_pii(clean_global_registry):
...
-                {'id:party2_ns/dataset/pii2': 3}, scenario['rules-party2']),
-            Site('site3', 'ddm', 'ddm_ns', {}, scenario['rules-ddm'])]
+            Site(name='site1', administrator='party1', namespace='party1_ns',
+                 stored_data={'id:party1_ns/dataset/pii1': 42},
+                 rules=scenario['rules-party1']),
+            Site(name='site2', administrator='party2', namespace='party2_ns',
+                 stored_data={'id:party2_ns/dataset/pii2': 3},
+                 rules=scenario['rules-party2']),
+            Site(name='site3', administrator='ddm', namespace='ddm_ns',
+                 stored_data={'id:ddm_ns/software/combine': 42,
```
svenvanderburg: I've just added random numbers here. It seems weird to have to pass an integer here, the data passed here should at some point represent the actual computation step right?

proof_of_concept/local_workflow_runner.py
```diff
@@ -174,6 +179,13 @@ def _get_step_inputs(
...
+        #  but should be retrieved in some way
+        data, metadata = self._ddm_client.retrieve_data(store_id='site3-store',
+                                                        name=compute_asset_name)
+        return ComputeAsset(compute_asset_name, data, metadata)
```
svenvanderburg: We could think about storing assets in the asset_store as a list of Dict[str, Asset]  (instead of 2 separate dicts with data and metadata), that would make this less clumsy. Let's discuss!

tests/test_scenarios.py
LourensVeen: I think it would be best to change this to Any for now, so that we can store an int for the data assets and a ComputeAsset object for the compute assets. Eventually it should become an Asset, but we cannot store data then because we don't have a DataAsset class yet. Ultimately it may even become a simple string that identifies a container image...

proof_of_concept/local_workflow_runner.py
svenvanderburg: We decided to make a hardcoded solution for now (that works for all tests) i.e. have a ddm which acts as a central repository of software.

svenvanderburg: And then solve this in https://dev.azure.com/nlesc-team-flow/Team%20FLOW/_workitems/edit/640

proof_of_concept/local_workflow_runner.py
svenvanderburg: Let's do this!

proof_of_concept/local_workflow_runner.py
svenvanderburg: Correction: instead of hardcoded store_id we can use `get_asset_location`.

svenvanderburg:
Make data asset-specific properties of MetaData optional

proof_of_concept/asset.py
```diff
@@ -1,21 +1,48 @@
 """Classes for describing assets."""
-from typing import Any, Dict
+from typing import Any, Dict, Union
+
+from proof_of_concept.workflow import Job
+
+
+class Metadata:
```
svenvanderburg, line 7: I moved Metadata here due to circular imports (and also makes sense tho have it here as it is not an interface)

proof_of_concept/ddm_site.py
```diff
@@ -68,9 +69,9 @@ def __init__(
         # Server side
         self.store = AssetStore(name + '-store', self._policy_evaluator)
         self._ddm_client.register_store(administrator, self.store)
-        for key, val in stored_data.items():
-            self.store.store(key, val, Metadata(Job.niljob(key), 'dataset'))
```
svenvanderburg, line 72: I moved this to the init of an Asset

proof_of_concept/local_workflow_runner.py
```diff
@@ -161,19 +166,24 @@ def _get_step_inputs(
...
-                print('Metadata: {}'.format(metadata))
+                print('Metadata: {}'.format(asset.metadata))
             except KeyError:
                 print('Job at {} found input {} not yet available.'.format(
                     self._this_runner, data_key))
                 return None
 
         return step_input_data
 
+    def _retrieve_compute_asset(self, compute_asset_id: str) -> ComputeAsset:
```
svenvanderburg, line 183: I'm getting: 185: error: Incompatible return value type (got "Asset", expected "ComputeAsset") from mypy here. I have no clue how to tell python that the store can have any subclass of Asset, but that I am expecting a ComputeAsset here (which might be a shortcoming of my current implementation). @LourensVeen let's discuss this!

tests/test_scenarios.py
svenvanderburg: I fixed this by actually implementing the DataAsset class. This simplified things a lot.

proof_of_concept/local_workflow_runner.py
svenvanderburg: Implemented this, seems more logical to me now.

proof_of_concept/asset.py
```diff
@@ -1,21 +1,48 @@
...
+        self.job = job
+        self.item = item
 
 
 class Asset:
     """Asset, a representation of a computation or piece of data."""
 
-    def __init__(self, name: str, data: Any, metadata: Any):
+    def __init__(self, id: str, data: Any,
+                 metadata: Union[Metadata, None] = None):
```
LourensVeen: typings has an Optional[X] generic type which is defined as Union[X, None] but is more readable.

proof_of_concept/local_workflow_runner.py
LourensVeen: If you check this using `isinstance()`, then mypy is smart enough to deduce for each branch that it is the checked-for type. If it's not a ComputeAsset, we should raise, because someone submitted an invalid workflow.

proof_of_concept/asset.py
```diff
@@ -1,21 +1,48 @@
...
         """Constructor.
 
-        TODO: typehint Metadata class for metadata argument, problematic
-            due to circular imports
         Args:
-            name: Name of the asset
+            id: Name of the asset
             data: Data related to the asset
             metadata: Metadata related to the asset
```
LourensVeen: Should document what a None argument means here.

proof_of_concept/local_workflow_runner.py
svenvanderburg: Fixed!

PR #15, opened 2020-08-21
Add logging

svenvanderburg:
Add logging to the project. 
Replace most print statements with logger.info statements (sometimes logger.warning)
Update logging statements where necessary
Add new logging statements where necessary

I think we log enough for now, if we need more logging for any reason we can always add it.

### DIFF
--- a/proof_of_concept/asset_store.py
+++ b/proof_of_concept/asset_store.py
@@ -1,10 +1,13 @@
 """Asset stores store data and compute assets."""
+import logging
 from typing import Dict
 
 from proof_of_concept.asset import Asset
 from proof_of_concept.definitions import IAssetStore
 from proof_of_concept.permission_calculator import PermissionCalculator
 from proof_of_concept.policy import PolicyEvaluator
+
+logger = logging.getLogger(__file__)
 
 
 class AssetStore(IAssetStore):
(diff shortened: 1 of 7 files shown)


proof_of_concept/local_workflow_runner.py
```diff
@@ -177,8 +177,8 @@ def _get_step_inputs(
                     self._this_runner, data_key))
                 logger.info('Metadata: {}'.format(asset.metadata))
             except KeyError:
-                logger.info(f'Job at {self._this_runner} found input'
-                            f'{data_key} not yet available.')
+                logger.warning(f'Job at {self._this_runner} found input'
```
LourensVeen: I think info is more appropriate here, this is normal expected behaviour. It just means that the previous step hasn't finished yet.

proof_of_concept/asset_store.py
```diff
@@ -53,17 +53,19 @@ def retrieve(self, asset_id: str, requester: str
             KeyError: If no asset with the given id is stored here.
 
         """
-        logger.info(f'{self} servicing request from {requester} for data: '
-                    f'{asset_id}, ')
+        logger.info(f'{self}: servicing request from {requester} for data: '
```
LourensVeen, line 56: Can we discuss f-strings and support for older Python versions?

proof_of_concept/local_workflow_runner.py
svenvanderburg: Updated

proof_of_concept/asset_store.py
svenvanderburg: We discussed that we will stick to f-strings for now, and therefore not support older python versions.

PR #16, opened 2020-08-24
Replace global registry request API with replication

LourensVeen:
The global registry had a direct API for discovering other sites and getting the location of their services. That doesn't scale. This pull request replaces that system with a replication-based solution using the existing replication system, which should work better for a small-data high-traffic data store.

### DIFF
--- a/proof_of_concept/registry.py
+++ b/proof_of_concept/registry.py
@@ -229,11 +229,9 @@ class Registry:
             site_name: The site this store is located at.
             store: The data store to register.
         """
-        for o in self._store.objects():
-            if isinstance(o, AssetStoreDescription) and o.store == store:
-                raise RuntimeError(
-                        'There is already a store called {}'.format(
-                            store.name))
+        if self._in_store(AssetStoreDescription, 'store', store):
+            raise RuntimeError(
+                    'There is already a store called {}'.format(store.name))
 
         site = self._get_object(SiteDescription, 'name', site_name)
         if site is None:
--- a/proof_of_concept/registry.py
+++ b/proof_of_concept/registry.py
@@ -175,8 +175,7 @@ class Registry:
             public_key: Public key of this party.
         """
         if self._in_store(PartyDescription, 'name', name):
-            raise RuntimeError(
-                    'There is already a party called {}'.format(name))
+            raise RuntimeError(f'There is already a party called {name}')
 
         party_desc = PartyDescription(name, public_key)
         self._store.insert(party_desc)
@@ -191,12 +190,11 @@ class Registry:
 
         """
         if self._in_store(SiteDescription, 'name', name):
-            raise RuntimeError(
-                    'There is already a site called {}'.format(name))
+            raise RuntimeError(f'There is already a site called {name}')
 
         admin = self._get_object(PartyDescription, 'name', admin_name)
         if admin is None:
-            raise RuntimeError('Party {} not found'.format(admin_name))
+            raise RuntimeError(f'Party {admin_name} not found')
 
         site_desc = SiteDescription(name, admin)
         self._store.insert(site_desc)
@@ -213,11 +211,11 @@ class Registry:
         """
         if self._in_store(RunnerDescription, 'runner', runner):
             raise RuntimeError(
-                    'There is already a runner called {}'.format(runner.name))
+                    f'There is already a runner called {runner.name}')
 
         site = self._get_object(SiteDescription, 'name', site_name)
         if site is None:
-            raise RuntimeError('Site {} not found'.format(site_name))
+            raise RuntimeError(f'Site {site_name} not found')
 
         runner_desc = RunnerDescription(site, runner)
         self._store.insert(runner_desc)
@@ -230,12 +228,11 @@ class Registry:
             store: The data store to register.
         """
         if self._in_store(AssetStoreDescription, 'store', store):
-            raise RuntimeError(
-                    'There is already a store called {}'.format(store.name))
+            raise RuntimeError(f'There is already a store called {store}')
 
         site = self._get_object(SiteDescription, 'name', site_name)
         if site is None:
-            raise RuntimeError('Site {} not found'.format(site_name))
+            raise RuntimeError(f'Site {site_name} not found')
 
         store_desc = AssetStoreDescription(site, store)
         self._store.insert(store_desc)
@@ -252,16 +249,16 @@ class Registry:
             server: The data store to register.
         """
         if self._in_store(PolicyServerDescription, 'server', server):
-            raise RuntimeError('The server is already registered')
+            raise RuntimeError(f'Server {server} is already registered')
 
         site = self._get_object(SiteDescription, 'name', site_name)
         if site is None:
-            raise RuntimeError('Site {} not found'.format(site_name))
+            raise RuntimeError(f'Site {site_name} not found')
 
         namespace = self._get_object(
                 NamespaceDescription, 'name', namespace_name)
         if namespace is None:
-            raise RuntimeError('Namespace {} not found'.format(namespace_name))
+            raise RuntimeError(f'Namespace {namespace_name} not found')
 
         server_desc = PolicyServerDescription(site, namespace, server)
         self._store.insert(server_desc)
--- a/proof_of_concept/registry.py
+++ b/proof_of_concept/registry.py
@@ -158,7 +158,7 @@ class Registry:
     """
     def __init__(self) -> None:
         """Create a new registry."""
-        self._assets = dict()           # type: Dict[str, str]
+        self._asset_locations = dict()           # type: Dict[str, str]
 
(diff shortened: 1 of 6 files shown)


proof_of_concept/registry.py
```diff
@@ -5,194 +5,279 @@
...
-            server: The data store to register.
+            name: Name of the site.
+            admin: Party which administrates this site.
+
         """
-        if server in self._policy_servers:
-            raise RuntimeError('This server is already registered')
-        self._policy_servers[namespace] = server
+        self.name = name
+        self.admin = admin
```
svenvanderburg, line 78: Is `admin` something different than `owner` (i.e. in `NamespaceDescription`)?

proof_of_concept/registry.py
```diff
@@ -5,194 +5,279 @@
...
 
         Args:
-            name: The name of the runner to look up.
+            name: Name of the party.
+            namespace: ID namespace owned by this party.
+            public_key: Public key of this party.
+        """
+        for i in self._store.objects():
+            if isinstance(i, PartyDescription) and i.name == name:
+                raise RuntimeError('There is already a party called {}'.format(
```
svenvanderburg: I would raise ValueError here

proof_of_concept/registry.py
```diff
@@ -5,194 +5,279 @@
...
+            admin_name: Party administrating this site.
 
-        Raises:
-            KeyError: If no runner with the given name is regstered.
         """
-        return self._runner_admins[name]
+        for o in self._store.objects():
+            if isinstance(o, SiteDescription) and o.name == name:
+                raise RuntimeError('There is already a site called {}'.format(
+                    name))
```
svenvanderburg: This pattern is repeated a lot, maybe you can replace it with a method:

```python
def _object_exists(name: str, type: Any):
    for o in self._store.objects():
        if isinstance(o, type) and o.name == name:
             return True
    return False
```

proof_of_concept/registry.py
```diff
@@ -5,194 +5,279 @@
...
-        Return:
-            A list of all registered policy servers and their
-                    namespaces.
+        for o in self._store.objects():
+            if isinstance(o, NamespaceDescription):
+                if o.name == namespace_name:
+                    namespace = o
+                    break
+        else:
+            raise RuntimeError('Namespace {} not found'.format(namespace_name))
```
svenvanderburg: It would be clearer to me to replace this with a method `_get_object`, similar to `_object_exists` (see my comment on line 194). With this you could also simplify a lot of code in the ddm client it seems.

proof_of_concept/registry.py
svenvanderburg: Maybe you can define this in the CanonicalStore class.

proof_of_concept/registry.py
svenvanderburg: Maybe in the `CanonicalStore` class

proof_of_concept/registry.py
```diff
@@ -5,194 +5,279 @@
...
 
-        Return:
-            The runner with that name.
+    Registers runners, stores, and assets. In a real system, runners
+    and stores would be identified by a URL, and use the DNS to
+    resolve. For now the registry helps with this.
+    """
+    def __init__(self) -> None:
+        """Create a new registry."""
+        self._assets = dict()           # type: Dict[str, str]
```
svenvanderburg: I would rename this to something like:
```suggestion
        self._asset_locations = dict()           # type: Dict[str, str]
```

proof_of_concept/registry.py
```diff
@@ -5,194 +5,279 @@
...
 
-        Raises:
-            KeyError: If no runner with the given name is
-                    registered.
-        """
-        return self._runners[name]
+        self._archive = ReplicableArchive[RegisteredObject]()
```
svenvanderburg: The archive is only needed to setup the store and replication_server, so maybe not assign it to `self`?

svenvanderburg (changes requested):
I have added some comments to the code, feel free to ignore my comments where you have good reasons to do something like you did it ;)

In general I want to mention that for me this is all pretty complicated to understand, probably because it is very low-level stuff and I am not used to that. I can sort of grasp the idea and look at the code in the PR and have an opinion about that code. But overall I do still not understand how this whole replication thing works. I think you could take 2 conclusions from this:

1. Collaborators/users like me that are not experienced in low-level code like this might get very confused by this part of the program. Not sure if that's a problem, but good to note I guess.
2. Related: It makes sense to abstract away the replication system as much as possible. Just like when I talk to a SQL database I just want to select/insert some data and not worry about the underlying low-level operations that are taking place related to replication. I have hinted a bit to this in my comments, but maybe you could think of more ways to do this. A separate repo might be a bridge too far, but it might not hurt to think how to make this sort of standalone and have a more user-friendly API to it (i.e. that would also benefit this project IMO)

I hope this helps :)

LourensVeen:
Here's a general answer to the general comment at the top: believe me, I know :-). This is systems programming, and it's quite a bit more complicated than writing script to analyse data, which is what much of scientific programming comes down to. But the goal of this project is to build a secure distributed scalable infrastructure to build applications on top of which are then controlled from such scripts, and that entails doing this kind of stuff.

The interface is not that complicated: on the server side there is an `insert()` and a `delete()` method, and the replica gives you the current set of objects and has an `update()` function to make sure it's up to date enough. I don't think I can make it any easier. Maybe in the construction of the canonical side, you need multiple objects now and that could be reduced if we make `archive` an optional argument to the `CanonicalStore` constructor. But `ReplicableArchive` is really just a placeholder for a relational database, if you have one of those then you'd be passing a connection string or something.

proof_of_concept/registry.py
LourensVeen: Yes, I had a software-as-a-service solution in mind here, where a company would offer DDM sites in the cloud. In that case the admin is this company (they control the infrastructure) but the owner is their customer/tenant.

proof_of_concept/registry.py
LourensVeen: Good point!

proof_of_concept/registry.py
LourensVeen: I considered it, but we don't know in general how the user wants to determine whether two objects are identical. In this case by the `name` attribute, but that's just how this class happens to work. You could have the user pass a lamba function, but that quickly becomes harder to understand than having a bit of repetition, IMHO. I'll have another look at it though, see if I can improve it, you're right that it's not optimal.

proof_of_concept/registry.py
LourensVeen: I agree, although then `for a in self._asset_locations` will iterate through asset names, which may be confusing. But it's conceptually a dictionary of locations indexed by name, so that is a better name. I'll change it.

proof_of_concept/registry.py
LourensVeen: Good point, will change.

proof_of_concept/registry.py
LourensVeen: Sorry, I've changed my mind :-). The Python docs say that `ValueError` should be raised if an "inappropriate" value is passed to a function. There are a few subclasses which define errors that are raised when one tries to process a bytes string as Unicode, but it doesn't form a valid UTF-8 sequence (or so I imagine :-)). So in that case, the contents of the variable are invalid.

Here, if you try to register a party twice, the name of the party is not invalid (presumably it was fine when it was first registered). It's just that the system is in a state in which the requested operation is impossible. So unless we define a subclass of RuntimeError for this specifically, I think RuntimeError is actually better.

proof_of_concept/registry.py
LourensVeen: Okay, I managed, although it does use a bit of introspection. But that's Pythonic and it looks better. What do you think?

proof_of_concept/registry.py
LourensVeen: I've added this now, but locally. I think it'd be good to refactor this into `CanonicalStore`, or into `Replica`, or maybe both. But that's probably better as as separate issue?

proof_of_concept/registry.py
LourensVeen: Done

proof_of_concept/registry.py
LourensVeen: Done.

proof_of_concept/registry.py
svenvanderburg: I would still raise ValueError here (but maybe that is because people often misuse it). But I  am totally fine with RuntimeError! :)

proof_of_concept/registry.py
svenvanderburg: This looks very nice actually (I guess I like Python and better looks)! Great work 👍

proof_of_concept/registry.py
svenvanderburg: Check, for now this looks good!

PR #17, opened 2020-08-26
Add asset_store_server REST api

svenvanderburg:
Replaces the old `AssetStore` class with an asset store server microservice that is accessed through a REST API. The rest of the program now talks to `AssetStoreClient`, which handles setting up the server in background threads and provides the same interface as the old `AssetStore` class.


proof_of_concept/asset_store_server/client.py
```diff
@@ -0,0 +1,82 @@
...
+        """Retrieves an asset.
+
+        Args:
+            asset_id: ID of the asset to retrieve.
+            requester: Name of the party making the request.
+
+        Return:
+            The asset object with asset_id.
+
+        TODO: We now solve permissions on the client side, which of course
```
svenvanderburg: I propose to do this in a separate PR and devops item.

proof_of_concept/asset.py
```diff
@@ -1,58 +1,66 @@
...
+            item (str): The item in the job's workflow corresponding to
+                    this asset.
         """
         self.job = job
         self.item = item
 
 
-class Asset:
-    """Asset, a representation of a computation or piece of data."""
+class Asset(Model):
```
svenvanderburg, line 32: I decided to just stick to Asset and not care about whether it is a data or compute asset. This seems to work flawlessly, not sure if it adheres to theoretical concepts, @LourensVeen ?

proof_of_concept/asset_store_server/run_api.py
```diff
@@ -0,0 +1,77 @@
...
+import connexion
+from flask import Flask
+from flask_injector import FlaskInjector
+
+from proof_of_concept.asset_store_server.store import AssetStore
+
+logger = logging.getLogger(__file__)
+
+
+def _configure_dependency_injection(flask_app: Flask, store: AssetStore
```
svenvanderburg: For me this is easiest to setup because I am used to dependency injection pattern, but I am curious whether you find this confusing.

svenvanderburg:
@LourensVeen There is still some mypy errors, I would like your help to fix those (because I am pulling my hair out and I know that you will probably have a good solution in 1 sec).

proof_of_concept/asset.py
LourensVeen: Good question. In the end they will all be containers, but they're used in different ways, and need to be implemented in different ways. That's outside of this code though. We do treat them differently in that we `run()` a ComputeAsset but never a DataAsset. I guess we will `start()` and `stop()` a DataAsset instead.

We could have separate subclasses, but we could also put all those methods on the Asset class. I don't really see a clear conceptual advantage to either. Is it two different kinds of things, or is it the same kind of thing (a container) that gets used in different ways? Let's keep Asset for now, it's the simplest.

proof_of_concept/asset_store_server/__init__.py
```diff
@@ -0,0 +1,2 @@
+"""Asset store server."""
+from .client import AssetStoreClient
```
LourensVeen, line 2: Could we do the full module path here please, instead of a relative one? I think it's best practice, although I forgot why :-).

proof_of_concept/asset_store_server/client.py
```diff
@@ -0,0 +1,85 @@
...
+
+    def __init__(self, name: str, policy_evaluator: PolicyEvaluator,
+                 host: str = 'http://127.0.0.1', port: int = 5000) -> None:
+        """Create a new AssetStoreClient."""
+        self.name = name
+        self._policy_evaluator = policy_evaluator
+        self._permission_calculator = PermissionCalculator(policy_evaluator)
+        self._port = port
+        self._api_url = f'{host}:{port}'
+        self._setup_server()
```
LourensVeen, line 28: Do I understand right that this starts an HTTP server? Isn't the client supposed to connect to a server, rather than create one? It would make more sense to me to have AssetStoreServer and AssetStoreClient classes, but maybe I'm missing something?

proof_of_concept/workflow.py
```diff
@@ -63,28 +88,33 @@ def __init__(
...
-        self.steps = dict()     # type: Dict[str, WorkflowStep]
-
-        for step in steps:
-            self.steps[step.name] = step
-
+        self.steps = steps
         self._validate()
 
+    @property
+    def steps_dict(self) -> Dict[str, WorkflowStep]:
```
LourensVeen, line 99: I'm not sure why it's an improvement to recalculate the steps dictionary every time we need it?

proof_of_concept/asset.py
```diff
@@ -1,38 +1,51 @@
...
-class Metadata:
-    """Stores metadata for stored assets.
+class Metadata(Model):
+    """Stores metadata for stored assets."""
+    swagger_types = {
+        'job': Job,
+        'item': str
+    }
 
-    Attributes:
```
LourensVeen, line 10: Why remove this documentation?

LourensVeen:
I would like to have the current design changed a bit, because the client and server side are a bit mixed up and that makes things complicated. Also, I'm not sure how we could have more than one `AssetStore` in a single program, which at least in the short term we're doing.

I think the following would work, but would appreciate feedback :-).

- [ ] Class AssetStore has a Flask app, policy evaluator, permission calculator and a db dict, and PE/PC/DB are stored in the app's config member. It's a server, not a client.
- [ ] The request handlers use `current_app` to access those objects. `retrieve` looks pretty much the same, except it is now a free function bound to an endpoint and gets the PE/PC/DB objects from the app config.
- [ ] Class AssetStore has a `get_endpoint()` which returns the URL to its REST endpoint.
- [ ] This endpoint is retrieved by `DDMSite` and sent to the Registry instead of the AssetStore object itself.
- [ ] The Registry is modified to store string with URLs rather than AssetStore objects.
- [ ] The `DDMClient` class's `retrieve_asset()` method gets this endpoint for the named store, instantiates a REST client for it, and uses it to retrieve the asset.

LourensVeen:
Closing this as it was superseded by the falcon-based version. Sorry, and thanks!

PR #19, opened 2020-09-17
Registry registers sites

LourensVeen:
This modifies the Registry to register sites, with optional runners, asset stores and policy servers, rather than registering those components individually. The idea is that this lets us have a single REST endpoint per site, rather than three, which simplifies things like firewalls and packaging, but comes at the cost of having an API with optional bits. I think that that's a good trade-off, so here we are.

### DIFF
--- a/proof_of_concept/ddm_site.py
+++ b/proof_of_concept/ddm_site.py
@@ -8,6 +8,7 @@ from cryptography.hazmat.primitives.asymmetric import rsa
 from proof_of_concept.asset import Asset
 from proof_of_concept.asset_store import AssetStore
 from proof_of_concept.ddm_client import DDMClient
+from proof_of_concept.definitions import PartyDescription
 from proof_of_concept.local_workflow_runner import LocalWorkflowRunner
 from proof_of_concept.policy import PolicyEvaluator, Rule
 from proof_of_concept.policy_replication import PolicySource
@@ -55,8 +56,8 @@ class Site:
                 backend=default_backend())
 
         self._ddm_client.register_party(
-                self.administrator,
-                self._private_key.public_key())
+                PartyDescription(
+                    self.administrator, self._private_key.public_key()))
 
         # Policy support
         self._policy_archive = ReplicableArchive[Rule]()
--- a/proof_of_concept/ddm_site.py
+++ b/proof_of_concept/ddm_site.py
@@ -22,7 +22,7 @@ logger = logging.getLogger(__file__)
 class Site:
     """Represents a single DDM peer installation."""
     def __init__(
-            self, name: str, administrator: str,
+            self, name: str, owner: str,
             namespace: str, stored_data: List[Asset],
             rules: List[Rule]) -> None:
         """Create a Site.
@@ -31,7 +31,7 @@ class Site:
 
         Args:
             name: Name of the site
-            administrator: Party which administrates this site.
+            owner: Party which owns this site.
             namespace: Namespace used by this site.
             stored_data: Data sets stored at this site.
             rules: A policy to adhere to.
@@ -39,10 +39,11 @@ class Site:
         """
         # Metadata
         self.name = name
-        self.administrator = administrator
+        self.owner = owner
+        self.administrator = owner
         self.namespace = namespace
 
-        self._ddm_client = DDMClient(administrator)
+        self._ddm_client = DDMClient(self.administrator)
 
         # Register party with DDM
         self._private_key = rsa.generate_private_key(
@@ -51,11 +52,8 @@ class Site:
                 backend=default_backend())
 
         self._ddm_client.register_party(
-                self.administrator, self.namespace,
+                self.administrator,
                 self._private_key.public_key())
-
-        # Register site with DDM
-        self._ddm_client.register_site(self.name + '-site', administrator)
 
         # Policy support
         self._policy_archive = ReplicableArchive[Rule]()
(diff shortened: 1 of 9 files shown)


proof_of_concept/ddm_site.py
```diff
@@ -31,18 +31,19 @@ def __init__(
...
+            owner: Party which owns this site.
             namespace: Namespace used by this site.
             stored_data: Data sets stored at this site.
             rules: A policy to adhere to.
 
         """
         # Metadata
         self.name = name
-        self.administrator = administrator
+        self.owner = owner
```
svenvanderburg, line 43: If owner == administrator, what's the point of having a distinction?

proof_of_concept/registry.py
```diff
@@ -38,115 +38,61 @@ def __init__(self, name: str, public_key: RSAPublicKey) -> None:
...
-        Args:
-            site: The site at which this server is located.
-            namespace: The namespace governed by this server.
-            server: The policy server.
-
-        """
-        self.site = site
-        self.namespace = namespace
-        self.server = server
+_ReplicatedClass = TypeVar('_ReplicatedClass', bound=RegisteredObject)
```
svenvanderburg: Maybe factor this out into a method `_validate()`

proof_of_concept/ddm_site.py
```diff
@@ -22,7 +22,7 @@
 class Site:
```
svenvanderburg, line 23: Both in `register_site` and `SiteDescription` the store, runner, namespace, and policy_server are optional. But not for a Site object. I guess this is a choice that for now we just want to test with sites that have everything?

proof_of_concept/registry.py
svenvanderburg: Additionally, I was wondering whether this validation lives here, and not under Site construction or when registering a site. (naive question)

svenvanderburg (approved):
Great work @LourensVeen! I agree that this is a trade-off, but I also agree that having a single REST API per site makes life a lot easier (and the benefits are higher than the costs here). Also I think this simplifies the code a lot! 

I've added some comments, but they are mainly to double check whether you thought everything through, I don't really expect any changes.

proof_of_concept/ddm_site.py
LourensVeen: The owner owns the namespace for this site (i.e. has the right to determine what to do with the assets in the namespace), while the administrator does the technical system administration of the site, and is therefore the party you need to trust in order to give access to your assets.

For now owners and administrators are the same party, but the idea here is that there may be providers offering a DDM-as-a-Service, in which case their customers would be owners, but the provider would be the administrator, so I'd like to keep them separate here so as to make this a more realistic POC.

proof_of_concept/registry.py
LourensVeen: It's not factoring out if there's only one occurrence :-). It would add two extra lines of code and an extra function, with the only benefit that you make it explicit that this is validating the object, but I think that's clear enough as is.

The reason I'm checking this here is that it simply makes no sense to have a namespace without a policy server for it or vice versa. So this is a class invariant of `SiteDescription`, and those should be checked in the constructor to ensure that there are no invalid instances around. Of course input validation should be done in any REST API version of  this, but that's a separate issue I think.

Come to think of it, we could actually have the user submit a `SiteDescription` object in `Registry.register_site()`, instead of a bunch of separate fields. I'd like to remove the `ReplicatedObject` base class then though, since it's an implementation detail of the Registry and should not be in a class used outside of it. It did occur to me some time ago that you can probably use a Union rather than having a base class, so that should be possible. I'll have a look; maybe it makes sense to do the same for parties then as well, and it may provide a cleaner base to start the REST API from.

As for `Site`, that's a separate class which represents the actual site, where `SiteDescription` represents a digital document in the Registry which describes a site. The `Site` constructor creates the policy server, so it should be correct by construction.

proof_of_concept/ddm_site.py
LourensVeen: Yes, eventually we may make a `Site` with optional components, but for now each site has everything for convenience.

proof_of_concept/registry.py
svenvanderburg: Making it more explicit then (I would still call that factoring out). I guess this is very subjective though so feel free to leave it like this! 👍 

Indeed submitting a `SiteDescription` object to `register_site` makes a lot of sense!

And regarding `Site`, indeed the current constructor produces a valid `Site` so no need to validate.

proof_of_concept/ddm_site.py
svenvanderburg: Cool! Maybe add a comment with this info?

LourensVeen: Yes, good idea.

proof_of_concept/registry.py
LourensVeen: I'm not sure if the term really comes from there, but I think about the algebraic transformation `a * x + a * y => a * (x + y)`, where there's a common factor `a` that is taken out of the equation and moved to the front, removing duplication. I'm not sure whether Python code is an algebraic ring however :-).

I've moved the `PartyDescription` and `SiteDescription` to definitions.py now, and use them when registering. This also changes `SiteDescription` to refer to the owner and the admin by name, as opposed to having a pointer to an object. That needs to happen anyway for serialisation, and it only adds one lookup in `DDMClient`, so that works out fine.

svenvanderburg: Sounds good! 👍

Issue #18, opened 2020-08-27
Identifying assets


LourensVeen:
In this system, applications and data are represented by assets. An asset will be implemented as a container image. Assets are added to the system by one of the participants (parties), by putting the asset into their AssetStore. Users can use an asset by referring to it in a workflow they submit. During planning of the workflow, permissions for the assets involved are looked up in a PolicyStore. During execution of the workflow, the asset will be retrieved from the AssetStore by a Runner, and instantiated into a container. Parties may cache copies of frequently used assets from elsewhere in their own AssetStore, so that they don't have to re-download them all the time.

Assets need to be identified somehow. We can distinguish here between primary assets, which come from outside the system and are put into it, and secondary assets, which have been created as a (final or intermediate) result of a workflow. Currently, primary assets are identified by an identifier of the form `id:<namespace>/<path>`. Both namespaces and paths may contain slashes, so splitting them requires a list of registered namespaces. This inconvenient, because we need to extract the namespace to e.g. figure out who owns the asset (it's the owner of the namespace).

For secondary assets, we currently use an id of the form `hash:<hash>`, where the hash is a hexadecimal string. The hash of a step output is generated by combining the hashes of the inputs and the compute asset, and hashing that in turn. As a result, if two submitted workflows produce an (intermediate) output in the same way, then those outputs will be identical (absent non-deterministic compute assets!) and will have the same hash. Also, this means that for any given workflow, the hashes for the (intermediate) results can be calculated from the workflow description itself. Thus, if subsequent steps are executed in different locations, then all we need is to send the whole workflow to both locations with a description of where each step is executed, and then they can each calculate the ids of the intermediate assets that they need to publish and retrieve, and find out where to get them.

This works, but there is one issue, which is that workflows have inputs which are primary assets. So a job consists of the workflow and the ids of the primary assets to feed into it. Unfortunately, we cannot determine from that id in which AssetStore it is stored, and so we cannot retrieve it.

Currently, this is worked around by registering the assets and their location in the global registry, but this doesn't scale. It is also undesirable from the perspective of a secure federated digital data marketplace, because it means that all assets are globally visible, while participants may want to share a particular asset with specific other participants, but not disclose its existence to anyone else.

[...]

PR #21, opened 2020-09-29
Attempt at REST APIs using falcon

LourensVeen:
Here's an attempt at adding REST APIs to the prototype using the falcon library, which allows implementing them in an object-oriented way, fitting the rest of the code better.

### DIFF
--- a/proof_of_concept/ddm_client.py
+++ b/proof_of_concept/ddm_client.py
@@ -15,8 +15,7 @@ from proof_of_concept.definitions import (
 from proof_of_concept.policy import (
         InAssetCollection, InPartyCollection, MayAccess, ResultOfComputeIn,
         ResultOfDataIn, Rule)
-from proof_of_concept.serialization import (
-        deserialize_asset, serialize, serialize_job_submission)
+from proof_of_concept.serialization import deserialize_asset, serialize
 from proof_of_concept.registry import global_registry, RegisteredObject
 from proof_of_concept.replication import ObjectValidator, Replica
 from proof_of_concept.replication_rest import ReplicationClient
@@ -233,8 +232,7 @@ class DDMClient:
         """
         site = self._get_site('name', site_name)
         if site is not None and site.runner:
-            data = serialize_job_submission(submission)
-            requests.post(f'{site.endpoint}/jobs', json=data)
+            requests.post(f'{site.endpoint}/jobs', json=serialize(submission))
         else:
             raise RuntimeError(f'Site or runner at site {site_name} not found')
 
--- a/proof_of_concept/ddm_client.py
+++ b/proof_of_concept/ddm_client.py
@@ -1,8 +1,10 @@
 """Functionality for connecting to other DDM sites."""
+from pathlib import Path
 import requests
 from typing import Any, List, Optional, Tuple
 
 from cryptography.hazmat.primitives.asymmetric.rsa import RSAPublicKey
+import ruamel.yaml as yaml
 
 from proof_of_concept.asset import Asset
 from proof_of_concept.definitions import (
@@ -10,7 +12,7 @@ from proof_of_concept.definitions import (
         Plan, SiteDescription)
 from proof_of_concept.serialization import serialize
 from proof_of_concept.registry import global_registry, RegisteredObject
-from proof_of_concept.replication import Replica
+from proof_of_concept.replication import Replica, ReplicationClient
 from proof_of_concept.workflow import Job, Workflow
 
 
@@ -24,7 +26,21 @@ class DDMClient:
 
         """
         self._party = party
+        # TODO: This will be passed in as an argument later.
         self._registry_endpoint = 'http://localhost:4413'
+
+        registry_api_file = Path(__file__).parent / 'registry_api.yaml'
+        with open(registry_api_file, 'r') as f:
+            registry_api_def = yaml.safe_load(f.read())
+
+        registry_client = ReplicationClient[RegisteredObject](
+                self._registry_endpoint + '/updates',
+                registry_api_def, 'RegistryUpdate', RegisteredObject)
+
+        # TODO: enable this when we can actually serialize runners
+        # and stores and policy servers.
+        # self._registry_replica = Replica[RegisteredObject](
+        #         registry_client)
         self._registry_replica = Replica[RegisteredObject](
                 global_registry.replication_server)
 
--- a/proof_of_concept/ddm_client.py
+++ b/proof_of_concept/ddm_client.py
@@ -113,6 +113,17 @@ class DDMClient:
                 self._registry_endpoint + '/parties',
                 json=serialize(description))
 
+    def deregister_party(self, name: str) -> None:
+        """Deregister a party with the Registry.
+
+        Args:
+            name: Name of the party to deregister.
+
+        """
+        r = requests.delete(f'{self._registry_endpoint}/parties/{name}')
+        if r.status_code == 404:
+            raise KeyError('Party not found')
+
     def register_site(self, description: SiteDescription) -> None:
         """Register a site with the Registry.
 
@@ -123,6 +134,17 @@ class DDMClient:
         requests.post(
                 self._registry_endpoint + '/sites',
                 json=serialize(description))
+
+    def deregister_site(self, name: str) -> None:
+        """Deregister a site with the Registry.
+
+        Args:
+            name: Name of the site to deregister.
+
+        """
+        r = requests.delete(f'{self._registry_endpoint}/sites/{name}')
+        if r.status_code == 404:
+            raise KeyError('Site not found')
 
     def register_asset(self, asset_id: str, site_name: str) -> None:
         """Register an Asset with the Registry.
--- a/proof_of_concept/ddm_client.py
+++ b/proof_of_concept/ddm_client.py
@@ -1,7 +1,8 @@
 """Functionality for connecting to other DDM sites."""
+import logging
 from pathlib import Path
 import requests
-from typing import Any, List, Optional, Tuple
+from typing import Any, Dict, List, Optional, Set, Tuple
 from urllib.parse import quote
 
 from cryptography.hazmat.primitives.asymmetric.rsa import RSAPublicKey
@@ -11,13 +12,55 @@ from proof_of_concept.asset import Asset
 from proof_of_concept.definitions import (
         IAssetStore, ILocalWorkflowRunner, IPolicyServer, PartyDescription,
         Plan, SiteDescription)
+from proof_of_concept.policy import (
+        InAssetCollection, InPartyCollection, MayAccess, ResultOfComputeIn,
+        ResultOfDataIn, Rule)
 from proof_of_concept.serialization import (
         deserialize_asset, serialize, serialize_job, serialize_plan)
 from proof_of_concept.registry import global_registry, RegisteredObject
-from proof_of_concept.replication import Replica
+from proof_of_concept.replication import ObjectValidator, Replica
 from proof_of_concept.replication_rest import ReplicationClient
 from proof_of_concept.validation import Validator
 from proof_of_concept.workflow import Job, Workflow
+
+
+logger = logging.getLogger(__name__)
+
+
+_PolicyClient = ReplicationClient[Rule]
+
+
+class RuleValidator(ObjectValidator[Rule]):
+    """Validates incoming policy rules by checking signatures."""
+    def __init__(self, ddm_client: 'DDMClient', namespace: str) -> None:
+        """Create a RuleValidator.
+
+        Checks that rules apply to the given namespace, and that they
+        have been signed by the owner of that namespace.
+
+        Args:
+            ddm_client: A DDMClient to use.
+            namespace: The namespace to expect rules for.
+        """
+        self._key = ddm_client.get_public_key_for_ns(namespace)
+        self._namespace = namespace
+
+    def is_valid(self, rule: Rule) -> bool:
+        """Return True iff the rule is properly signed."""
+        if isinstance(rule, ResultOfDataIn):
+            namespace = rule.data_asset[3:].split('.')[0]
+        elif isinstance(rule, ResultOfComputeIn):
+            namespace = rule.compute_asset[3:].split('.')[0]
+        elif isinstance(rule, MayAccess):
+            namespace = rule.asset[3:].split('.')[0]
+        elif isinstance(rule, InAssetCollection):
+            namespace = rule.asset[3:].split('.')[0]
+        elif isinstance(rule, InPartyCollection):
+            namespace = rule.collection[3:].split('.')[0]
+
+        if namespace != self._namespace:
+            return False
+        return rule.has_valid_signature(self._key)
 
 
 class DDMClient:
@@ -33,6 +76,7 @@ class DDMClient:
         # TODO: This will be passed in as an argument later.
         self._registry_endpoint = 'http://localhost:4413'
 
+        # Set up connection to registry
         registry_api_file = Path(__file__).parent / 'registry_api.yaml'
         with open(registry_api_file, 'r') as f:
             registry_api_def = yaml.safe_load(f.read())
@@ -41,20 +85,22 @@ class DDMClient:
 
         registry_client = ReplicationClient[RegisteredObject](
                 self._registry_endpoint + '/updates', registry_validator,
-                'RegistryUpdate')
+                'RegistryUpdate', 'RegisteredObject')
 
+        self._registry_replica = Replica[RegisteredObject](
+                registry_client, on_update=self._on_registry_update)
+
+        # Set up connections to sites
         site_api_file = Path(__file__).parent / 'site_api.yaml'
         with open(site_api_file, 'r') as f:
             site_api_def = yaml.safe_load(f.read())
 
         self._site_validator = Validator(site_api_def)
 
-        # TODO: enable this when we can actually serialize runners
-        # and stores and policy servers.
-        # self._registry_replica = Replica[RegisteredObject](
-        #         registry_client)
-        self._registry_replica = Replica[RegisteredObject](
-                global_registry.replication_server)
+        self._policy_replicas = dict()  # type: Dict[str, Replica[Rule]]
+
+        # Get initial data
+        self._registry_replica.update()
 
     def register_party(self, description: PartyDescription) -> None:
         """Register a party with the Registry.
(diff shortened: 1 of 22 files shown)


proof_of_concept/serialization.py
```diff
@@ -0,0 +1,290 @@
+"""(De)Serializes objects of various kinds to JSON."""
```
svenvanderburg, line 1: This seems like a lot of code to do (de)serialization. I kind of liked the [code from swagger](https://github.com/SecConNet/proof_of_concept/pull/17/files#diff-b9dbc7408f3c86f83c17dd179e786c09). Curious what you think!

proof_of_concept/registry_api.yaml
```diff
@@ -0,0 +1,207 @@
+openapi: "3.0.0"
```
svenvanderburg, line 1: I guess there's good reasons to not do this, but why instead of the /updates endpoint we have a GET method for the /parties and /sites endpoints? Then we could get rid of the whole replication code.

I guess the answer is scalability. I would favour simplicity over scalability for now, but I am guessing you think differently ;)

proof_of_concept/registry_api.yaml
```diff
@@ -0,0 +1,207 @@
...
+
+info:
+    version: 0.1.0
+    title: EFDDM Registry
+
+paths:
+  /parties:
```
svenvanderburg, line 8: Why is there no DELETE method for parties? (same question for sites)

proof_of_concept/serialization.py
svenvanderburg: Even if the (de)serialization code needs to be custom for each class, I would still prefer having a `from_json()` and `to_json()` method on the class itself.

proof_of_concept/registry_api.py
```diff
@@ -0,0 +1,111 @@
...
+        """
+        self.app = App()
+
+        registry_api_file = Path(__file__).parent / 'registry_api.yaml'
+        with open(registry_api_file, 'r') as f:
+            registry_api_def = yaml.safe_load(f.read())
+
+        deserializer = Deserializer(registry_api_def)
+
+        party_registration = PartyRegistration(registry, deserializer)
```
svenvanderburg: Now we have to manually do the routing and we see that we don't fully make use of the power of swagger + connexion. But to be honest, I really don't mind actually, this makes it much less magical.

svenvanderburg (commented):
This looks really nice @LourensVeen! I have looked at your changes from a birds-eye perspective. I kind of like falcon, I think you could do something like this (i.e. the object-oriented approach) in flask with class-based views and dependency injection, but this seems much more elegant. And I don't think we need any of the flask features that falcon might not have. So: nice find, I definitely see myself moving to falcon instead of flask for some cases, at least not always by default using flask without considering other options.

proof_of_concept/serialization.py
LourensVeen: Introspection makes the code more difficult to understand though, I consider that a bit of an advanced feature. And so far all our classes have something funny about them that would complicate this. I started implementing to_json() and from_json(), but got stuck on the replica updates (which are a generic class) because typings/mypy isn't quite advanced enough. In C++ this would work beautifully, but well...

So I dropped that after spending way too much time on it (if I can find some time, I'll go write a feature proposal for typings), and instead decided to keep the aspect oriented design, but split up validation and serialization. That makes things a bit cleaner and means we're now doing the same thing with less code.

proof_of_concept/registry_api.yaml
LourensVeen: Yes, it's clear that a solution where you're continuously querying other services over the internet for this kind of information doesn't scale, and I want my proof of concept to prove the concept without having to rely on hand waving as to how it translates into practice. This is also eScience after all...

proof_of_concept/registry_api.yaml
LourensVeen: Erm, because I haven't got around to implementing it :). Probably should...

proof_of_concept/serialization.py
LourensVeen: I also tried to use YAtiML (with new JSON-functionality) but that didn't work because ruamel.yaml and cryptography's RSAPublicKey are fundamentally incompatible, and then I implemented a mini-serialiser to do the same thing using introspection, but it ended up being more lines of code and using more advanced Python features because things are quite irregular (Signature base class, RSAPublicKey, timestamp, class hierarchies with disambiguation fields, there are very few straightforward struct-type classes, so you end up with so many exceptions that you may as well write it by hand).

proof_of_concept/registry_api.yaml
LourensVeen: Implemented this.

proof_of_concept/ddm_client.py
```diff
@@ -106,7 +208,19 @@ def retrieve_asset(self, site_name: str, asset_id: str
...
         if site is not None and site.store is not None:
-            return site.store.retrieve(asset_id, self._party)
+            safe_asset_id = quote(asset_id, safe='')
+            r = requests.get(
+                    f'{site.endpoint}/assets/{safe_asset_id}',
+                    params={'requester': self._party})
+            if r.status_code == 404:
+                raise KeyError('Asset not found')
+            elif not r.ok:
+                raise RuntimeError('Server error when retrieving asset')
```
svenvanderburg, line 223: Do we know for sure it is a server error (i.e. 500) or can it also be something else?

proof_of_concept/registry_api.yaml
svenvanderburg: Check.

proof_of_concept/registry_api.yaml
svenvanderburg: Nice!

proof_of_concept/replication_rest.py
```diff
@@ -0,0 +1,86 @@
+"""REST API handlers/clients for the replication system."""
```
svenvanderburg, line 1: Wouldn't it be more consistent with the other apis to call this file `replication_api.py`?

proof_of_concept/serialization.py
svenvanderburg: Seems like you had quite an adventure here! Sorry if I pushed you into a rabbithole...

By looking at the (de)serialization methods I agree that there are a lot of exceptions that would make something like [this](https://github.com/SecConNet/proof_of_concept/pull/17/files#diff-b9dbc7408f3c86f83c17dd179e786c09) not always work (this is what you call the 'introspection'-method right?).

But wouldn't it work on the cases that are not exceptions? I think you would still cover about 50% of the regular models? I guess you can also have existing solutions like [dataclasses-json](https://github.com/lidatong/dataclasses-json) and [jsonpickle](https://github.com/jsonpickle/jsonpickle) to do this for those 50% regular ones.

Maybe you can have some Mixin class `ToDict` that implements a `to_json()` and `from_json()` method for the regular 50% (e.g. SiteDescription) and overwrite those methods for the irregular 50% (e.g. PartyDescription)?

If you choose to keep the (de)serialization code as it is, than I would still suggest to move it to the class itself. So PartyDescription().to_json() instead of serialize(PartyDescription()). I would like that better because the serialization code is in the class itself, making it easier why we serialize in that way. But I guess that's just a matter of preference and this is also fine!

tests/conftest.py
```diff
@@ -3,16 +3,23 @@
...
 import logging
+from threading import Thread
 from unittest.mock import patch
+from wsgiref.simple_server import WSGIRequestHandler, WSGIServer
 
 from cryptography.hazmat.primitives.asymmetric import rsa
 from cryptography.hazmat.backends import default_backend
 import pytest
 
+# TODO: the below hack will go when the complete API is implemented,
```
svenvanderburg, line 14: The complete API is implemented right?

tests/conftest.py
```diff
@@ -22,6 +29,38 @@ def clean_global_registry():
         yield None
 
 
+class ReusingWSGIServer(WSGIServer):
+    """A simple WSGI server which allows reusing the port.
+
+    There is of course a good reason for the existence of TIME_WAIT,
```
svenvanderburg: Why the comment on TIME_WAIT? I don't see TIME_WAIT anywhere?

proof_of_concept/replication_rest.py
```diff
@@ -0,0 +1,86 @@
...
+        response.media = serialize(updates)
+
+
+class ReplicationClient(IReplicationSource[T]):
+    """Client for a ReplicationHandler REST endpoint."""
+    def __init__(
+            self, endpoint: str,
+            validator: Validator, update_type_tag: str, content_type_tag: str
+            ) -> None:
+        """Create a ReplicationClient.
```
svenvanderburg, line 50: It would maybe help to add examples of `update_type_tag` and `content_type_tag` strings to make this a bit less abstract.

proof_of_concept/registry_api.py
```diff
@@ -0,0 +1,168 @@
...
+            self._validator.validate('Site', request.media)
+            self._registry.register_site(
+                    deserialize_site_description(request.media))
+            response.status = HTTP_201
+            response.body = 'Created'
+        except ValidationError as e:
+            logger.error(f'Invalid site description {e}')
+            response.status = HTTP_400
+            response.body = 'Invalid request'
+        except RuntimeError as e:
```
svenvanderburg, line 110: Are you sure there are no other RuntimeErrors raised?

svenvanderburg (commented):
Great work @LourensVeen! This took quite some effort, and I am really happy this works now! I left some small comments, but most of it looks really good. Some general comments on the PR as a whole:
1. The number of modules under `proof_of_concept` is rising steadily, maybe it makes sense to add some hierarchy? For example put `registry.py`, `registry_api.py`, and `registry_api.yml` under `proof_of_concept/registry`.
2. This took me more than 2 hours to review, and I still have the feeling I might have missed some details. I really think it would be better if you try to break stuff down into smaller steps. If you don't like pushing suboptimal solutions to `develop` you can always create seperate PRs to merge into `new-rest-api`.
3. The only link between the swagger specs and the actual api is through the schema validation right? This is potentially error-prone, since making a small typo in the swagger spec will not result in any errors, right? I guess we should either be very careful with this, or test this in a way.
4. Another thought on this weak connection between the swagger spec and the actual API is: Why do we need the swagger spec at all? For public-facing APIs I really see the benefits of swagger, but since it is likely gonna be DDM-components talking to each other (correct me if I am wrong on this) I doubt if anyone will ever look at the .yml files.

tests/conftest.py
LourensVeen: TIME_WAIT is a state that a listening TCP port is put into after it is closed, preventing another application from reusing that port immediately. The reason for this is that clients may still be trying to connect to the now-shutdown server, and if the new server program uses the same port then it can intercept those connections, which is a potential security issue. However, since we don't want to wait a couple of minutes between tests, and there's no actual sensitive anything in there, I'm disabling this feature by setting option.

I'm not sure that this whole explanation should be in the code there though...

svenvanderburg: OK, thanks for the clarification! Someone who doesn't know anything about this (like me) will not know that the comment as it is now actually refers to the allow_reuse_address option. So maybe make that a bit more explicit?

proof_of_concept/ddm_client.py
LourensVeen: Well, either the asset is found and we return it (HTTP 200), or it hasn't been found (HTTP 404) and we give an error. It may be that the asset doesn't exist yet, but that is still a not found, in the future when I implement proper error handling for workflows there may be a "run failed, won't exist, here's the error message" response, but that doesn't exist yet (and will still be an HTTP 200). So if you get another message then something went wrong unexpectedly, which means there was a server error.

requests.Response.ok is False if the status code is 400 or larger. Since we can only get a 200, 404, or 500, and 404 has already been handled, there must be either a 500 series code signalling a server error, or an unexpected 400 series code, which since it shouldn't have been sent also means there was a server error.

proof_of_concept/registry_api.py
LourensVeen: No. But I'm letting myself be a bit more sloppy with error handling on account of this being a prototype. Having a bunch of more specific exception classes would be better (I made https://github.com/SecConNet/proof_of_concept/issues/23). On the other hand, this is Python. The standard library doesn't come with exception guarantees, so you can never really know which exceptions might occur. You need Java or C++ or something for that.

proof_of_concept/replication_rest.py
LourensVeen: Yes, or call the others _rest. And there are handlers and servers and clients and it's all a bit messy. I've made https://github.com/SecConNet/proof_of_concept/issues/24 for this to try to not make this PR even bigger.

proof_of_concept/replication_rest.py
LourensVeen: I think I have a better way now to solve this, inspired by replication system number 4. I'll put that in, I agree that this is messy.

proof_of_concept/serialization.py
LourensVeen: I'm not using data classes (which are Python 3.7+ anyway and I like to have my code work on all supported Python versions), and jsonpickle explicitly says it's a security hole (although if you validate first it may be okay, but I don't feel for digging into it).

I'm going to stick with the aspect oriented design, putting all the serialization stuff in a separate module. I'm doing the same thing for signatures on objects (see signable.py), and it just seems like the best way of keeping the complexity under control.

tests/conftest.py
LourensVeen: Ah, yes, but we're still registering assets with the registry, and that doesn't go through the REST API. Once the new identifiers are there (https://github.com/SecConNet/proof_of_concept/issues/18) it can go, but not yet...

proof_of_concept/serialization.py
LourensVeen: Okay, cleaned things up and it's looking nice now. Maybe a bit repetitive at first glance, but this is really the best solution, and a bit of repetition is also fine. DRY is not an iron law.

LourensVeen:
> 1. The number of modules under `proof_of_concept` is rising steadily, maybe it makes sense to add some hierarchy? For example put `registry.py`, `registry_api.py`, and `registry_api.yml` under `proof_of_concept/registry`.

Yes, I'm going to rearrange all the classes/modules/packages, will be #24.

> 2. This took me more than 2 hours to review, and I still have the feeling I might have missed some details. I really think it would be better if you try to break stuff down into smaller steps. If you don't like pushing suboptimal solutions to `develop` you can always create seperate PRs to merge into `new-rest-api`.

I've looked at it, and it seems that the PR contains quite a few unrelated improvements. Should have done those first I think, and be more careful about that. The above clean-up will help.

> 3. The only link between the swagger specs and the actual api is through the schema validation right? This is potentially error-prone, since making a small typo in the swagger spec will not result in any errors, right? I guess we should either be very careful with this, or test this in a way.

Yes, so if the OpenAPI spec does not match the data being transmitted then we will notice, but the actual endpoint isn't checked against it. Maybe a test could be added that auto-generates a client from the spec and then calls a server and tests that the semantics are right? But that means more things to change when the API is updated as well, and we have an end-to-end test. If I had an infinite amount of time I'd do it, but I think for now I'll leave it as-is.

> 4. Another thought on this weak connection between the swagger spec and the actual API is: Why do we need the swagger spec at all? For public-facing APIs I really see the benefits of swagger, but since it is likely gonna be DDM-components talking to each other (correct me if I am wrong on this) I doubt if anyone will ever look at the .yml files.

We could remove them, but then we'd have to do the validation in some other way, probably handwritten too. I like having them however, since one idea of something to come out of this project is an open standard for how such a DDM works. The YAML files would come in handy then. So I'm leaving them in for now.

proof_of_concept/registry_api.py
LourensVeen: Yeah. If I had the time, I'd write a code generator that generates this code from an OpenAPI spec. Next life...

tests/conftest.py
LourensVeen: Yeah, I wrote that comment with the intended users in mind, who are computer scientists in the field of network engineering...they know what I mean. I've reworded it, maybe it works better now for you as well?

proof_of_concept/replication_rest.py
LourensVeen: And they're gone. Problem here is that Python's type annotations aren't really suitable for generic programming. But I'm not going to rewrite in C++ :-). I've added `UpdateType` and `ReplicatedType` members to some of the generic replication-related classes, and create derived classes from them where those are filled in more specifically so the correct type is deserialised. Looks nicer I think.

proof_of_concept/registry_api.py
svenvanderburg: Ok, fair enough! (And indeed smart to create an issue for it)

proof_of_concept/replication_rest.py
svenvanderburg: Bueno

tests/conftest.py
svenvanderburg: Alright. After your rewording I still don't fully understand it (you still have to know what TIME_WAIT is), but at least it is clear in what direction I would need to dig up some context if I want to edit this code.

svenvanderburg:
1. Nice that you created a new issue for this!
2. Yeah, let's be aware of it in the future!
3. That would indeed be a nice solution, but I agree that it is not high priority. Maybe create a low-priority issue for this? Which you will probably skip forever, but if you have the time at some point it might be nice to already have the solution laid out for you there.
4. Yes for validation it is indeed nice. I agree on keeping them, you can always kick them out at some point if you feel they are more a worry than a help.

tests/conftest.py
svenvanderburg: Check

proof_of_concept/replication_rest.py
svenvanderburg: Cool

proof_of_concept/ddm_client.py
svenvanderburg: Yes, that makes sense.

proof_of_concept/definitions.py
```diff
@@ -144,14 +144,12 @@ def __repr__(self) -> str:
...
-    def get_updates_since(
-            self, from_version: Optional[int]
-            ) -> ReplicaUpdate[T]:
+    def get_updates_since(self, from_version: int) -> ReplicaUpdate[T]:
         """Return a set of objects modified since the given version.
 
         Args:
             from_version: A version received from a previous call to
-                    this function, or None to get an update for a
+                    this function, or 0 to get an update for a
```
svenvanderburg, line 172: Good change :)

PR #36, opened 2020-11-04
Issue 28: Factor PolicyServer replicas out of DDMClient

### LINKED ISSUE #28: Factor PolicyServer replicas out of DDMClient
Part of #27.

- [x] Add a parameter for a list of callbacks to DDMClient __init__(), will be called when the registry replica is updated
- [x] Move the replicas of the policy servers to PolicySource, and register it with the DDMClient in Site so that it gets updates

LourensVeen:
This implements #28. There's a slight change relative to the plan, because we were getting a chicken-and-egg problem with registering handlers, so now there's a `DDMClient.register_callback()` instead of passing it to the init.

### DIFF
--- a/proof_of_concept/ddm_site.py
+++ b/proof_of_concept/ddm_site.py
@@ -79,7 +79,8 @@ class Site:
             self._policy_store.insert(rule)
         self.policy_server = PolicyServer(self._policy_archive, 0.1)
 
-        self._policy_source = PolicySource(self._ddm_client)
+        self._policy_source = PolicySource(
+                self._ddm_client, self._site_validator)
         self._policy_evaluator = PolicyEvaluator(self._policy_source)
 
         # Server side
--- a/proof_of_concept/ddm_site.py
+++ b/proof_of_concept/ddm_site.py
@@ -1,9 +1,11 @@
 """This module combines components into a site installation."""
 import logging
+from pathlib import Path
 from typing import Any, Dict, List
 
 from cryptography.hazmat.backends import default_backend
 from cryptography.hazmat.primitives.asymmetric import rsa
+import ruamel.yaml as yaml
 
 from proof_of_concept.asset import Asset
 from proof_of_concept.asset_store import AssetStore
@@ -14,6 +16,7 @@ from proof_of_concept.local_workflow_runner import LocalWorkflowRunner
 from proof_of_concept.policy import PolicyEvaluator, Rule
 from proof_of_concept.policy_replication import PolicyServer, PolicySource
 from proof_of_concept.replication import CanonicalStore, ReplicableArchive
+from proof_of_concept.validation import Validator
 from proof_of_concept.workflow import Job
 from proof_of_concept.workflow_engine import GlobalWorkflowRunner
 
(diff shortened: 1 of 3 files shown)


LourensVeen:
Set this back to draft temporarily, there's one more clean-up I want to do here. Go ahead and review what's there so far if you want, it's not a big change.

proof_of_concept/policy_replication.py
```diff
@@ -1,44 +1,119 @@
...
+                        o.endpoint + '/rules/updates', self._site_validator)
+
+                key = self._ddm_client.get_public_key_for_ns(o.namespace)
+                validator = RuleValidator(o.namespace, key)
+                self._policy_replicas[o.namespace] = Replica[Rule](
+                        client, validator)
+
+        for o in deleted:
+            if isinstance(o, SiteDescription) and o.namespace:
+                del(self._policy_replicas[o.namespace])
```
svenvanderburg, line 108: Nothing happens when there is no `o.namespace`? Do we need to signal this in any way?

proof_of_concept/ddm_site.py
```diff
@@ -48,7 +51,15 @@ def __init__(
         self.administrator = owner
         self.namespace = namespace
 
-        self._ddm_client = DDMClient(self.name)
+        # Load API definitions
+        site_api_file = Path(__file__).parent / 'site_api.yaml'
+        with open(site_api_file, 'r') as f:
+            site_api_def = yaml.safe_load(f.read())
+
+        self._site_validator = Validator(site_api_def)
```
svenvanderburg, line 59: There is a lot of passing around of this site_validator objects which just depends on a static file, it almost feels like a constant. In the end we only use it to init a PolicyClient and a DDMClient. This was kind of confusing for me, because it was hard to track why we would need to pass around this thing that is only needed to validate API input/output. I think the declaration of it in DDMClient made sense, maybe we can declare it in PolicyClient as well? Or just have a module `validation.py` and move all validator declerations there?

svenvanderburg (approved):
Just 2 small comments. For the rest it looks good! (Although I find it incredibly difficult to wrap my head around it. Your documentation helps a lot though.)

proof_of_concept/policy_replication.py
LourensVeen: No, this is fine, it just means that this site doesn't have a PolicyServer, which can happen if it's a service provider for instance which only offers compute capacity and doesn't have any assets to contribute. However, if it doesn't have a PolicyServer, then we don't have a Replica for it either, so we cannot delete it.

proof_of_concept/ddm_site.py
LourensVeen: Hmm. I'm not generally a fan of singletons (and they're considered a bad idea in general as well), but the fact that it's basically just a wrapper around the YAML file may tip the balance here. I don't like duplicating the code, but having a static object in validation.py may make sense (and then later a rest/validation/ directory with the YAML files and validation.py), and you're right that lugging this thing around everywhere is also not optimal. Let me sleep on it for a bit :-).

proof_of_concept/policy_replication.py
svenvanderburg: Cool

proof_of_concept/ddm_site.py
svenvanderburg: I suggest maybe making a separate issue for it, I see that also in in the other PRs you use this pattern. And then you can  just take the time to sleep on it :)

svenvanderburg: Then you can go ahead and merge this.

LourensVeen: Some useful background: https://docs.microsoft.com/en-gb/archive/blogs/scottdensmore/why-singletons-are-evil

Issue #33, opened 2020-11-02
Clean up ...Server and ...Source terminology


LourensVeen:
Part of #24.

We have classes named ...Server for HTTP servers for REST endpoints, but also for sources of replica updates, and then there's IReplicationSource which is an interface for ReplicationServer, and IPolicySource which is used for dependency inversion between PolicySource and PolicyEvaluator. Let's clean that up.

ReplicationServer will be merged into CanonicalStore in #32, which will also get rid of PolicyServer and RegistryServer. SiteServer is an HTTP server and is fine. If we

- [x] Rename IPolicySource to IPolicyBase
- [x] Rename IPolicyServer to IPolicySource (it derives from IReplicationSource, so that's consistent)

and leave IReplicationSource as is, then ...Server will be an HTTP server, ...Source will be a source of replica updates, and ...Base will offer access to a set of objects.

LourensVeen:
While we're at it, we should rename the ...Api classes to ...RestApi, and any REST clients should be named ...RestClient.

PR #37, opened 2020-11-05
Split DDMClient into RegistryClient and PeerClient

### LINKED ISSUE #29: Split RegistryClient out of DDMClient
Part of #27.

One of the things DDMClient does is to manage a local replica of the registry. That should be a separate class, so refactor it out.

- [x] Has a RegistryReplica(Replica[RegisteredObject])
- [x] Methods for (de)registering parties and sites
- [x] get_public_key_for_ns(), list_sites_with_runners(), get_site_administrator()
- [x] Temporarily keep register_asset(), get_asset_location() until #18 is implemented
- [x] Take callbacks (for PolicySource, #28) for init and call them with updates
- [x] Modify Site to set this all up

DDMClient will keep asset registration, submitting jobs to sites, and getting assets from sites. It will get endpoints by site name from a RegistryReplica as needed.

LourensVeen:
This one is designed to apply on top of #36, so that needs to be merged first. This splits DDMClient into two classes, RegistryClient does registration with the global registry and contains a replica of same, while PeerClient is for connecting to other sites to submit jobs and retrieve assets. For now, RegistryClient also still contains the asset registration code, but that will be removed when the new Asset IDs are implemented (#18).

Closes #29.

### DIFF
--- a/proof_of_concept/ddm_client.py
+++ b/proof_of_concept/ddm_client.py
@@ -22,7 +22,7 @@ from proof_of_concept.validation import Validator
 from proof_of_concept.workflow import Job, Workflow
 
 
-class RegistryClient(ReplicationClient[RegisteredObject]):
+class RegistryRestClient(ReplicationClient[RegisteredObject]):
     """A client for the registry."""
     UpdateType = RegistryUpdate
 
@@ -56,7 +56,7 @@ class DDMClient:
 
         registry_validator = Validator(registry_api_def)
 
-        registry_client = RegistryClient(
+        registry_client = RegistryRestClient(
                 self._registry_endpoint + '/updates', registry_validator)
 
         self._registry_replica = Replica[RegisteredObject](
--- a/proof_of_concept/ddm_client.py
+++ b/proof_of_concept/ddm_client.py
@@ -36,19 +36,10 @@ class RegistryReplica(Replica[RegisteredObject]):
     pass
 
 
-class DDMClient:
-    """Handles connecting to global registry, runners and stores."""
-    def __init__(self, site: str, site_validator: Validator) -> None:
-        """Create a DDMClient.
-
-        Args:
-            site: The site at which this client acts.
-            site_validator: A validator for the Site REST API.
-
-        """
-        self._site = site
-        self._site_validator = site_validator
-
+class RegistryClient:
+    """Local interface to the global registry."""
+    def __init__(self) -> None:
+        """Create a RegistryClient."""
         # TODO: This will be passed in as an argument later.
         self._registry_endpoint = 'http://localhost:4413'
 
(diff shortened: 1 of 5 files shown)


proof_of_concept/ddm_client.py
```diff
@@ -56,10 +52,10 @@ def __init__(self, site: str, site_validator: Validator) -> None:
 
         registry_validator = Validator(registry_api_def)
 
-        registry_client = RegistryClient(
+        registry_client = RegistryRestClient(
```
svenvanderburg, line 55:
```suggestion
        registry_rest_client = RegistryRestClient(
```
To make it less confusing.

proof_of_concept/ddm_client.py
```diff
@@ -22,7 +22,7 @@
 from proof_of_concept.workflow import Job, Workflow
 
 
-class RegistryClient(ReplicationClient[RegisteredObject]):
+class RegistryRestClient(ReplicationClient[RegisteredObject]):
     """A client for the registry."""
```
svenvanderburg, line 26: I think this makes sense (I cannot really think of a better name, apart from maybe `RegistryApiClient`, but that comes down to the same thing). Maybe you can make it a bit more explicit in the docstring that this is only for talking to the REST registry service.

svenvanderburg: To make a better distinction with `RegistryClient` class.

proof_of_concept/ddm_client.py
```diff
@@ -166,17 +162,82 @@ def list_sites_with_runners(self) -> List[str]:
...
+    """Handles connecting to other sites' runners and stores."""
+    def __init__(
+            self, site: str, site_validator: Validator,
+            registry_client: RegistryClient
+            ) -> None:
+        """Create a DDMClient.
+
+        Args:
+            site: The site at which this client acts.
+            site_validator: A validator for the Site REST API.
```
svenvanderburg, line 224: See my comment in #36 on passing around validators.

svenvanderburg (commented):
Nice to split this! I've left a few minor comments, otherwise it is good to go!

proof_of_concept/ddm_client.py
LourensVeen: I agree, but can we do this as part of #33? I think it makes more sense there.

proof_of_concept/ddm_client.py
LourensVeen: And likewise ;-)

proof_of_concept/ddm_client.py
LourensVeen: Agreed, #33

proof_of_concept/ddm_client.py
svenvanderburg: Sure!

PR #38, opened 2020-11-05
Make Plan refer to steps by name

### LINKED ISSUE #30: Have Plan refer to steps by name rather than by pointer
Part of #24.

It seems like the name is what we need anyway everywhere this is used, and not having to reconnect makes the replication simpler as well, so let's change this.

LourensVeen:
This applies on top of #37, so that should be merged first.

Closes #30.


svenvanderburg (approved):
Looks good!

PR #39, opened 2020-11-05
Rename workflow components

LourensVeen:
Closes #31.

PR #40, opened 2020-11-05
Merge ReplicationServer into CanonicalStore

### LINKED ISSUE #32: Merge ReplicationServer into CanonicalStore
Part of #24.

The whole point of a CanonicalStore is that you can create replica updates from it, which is what ReplicaServer does. The current architecture mimics a system where you have a relational database (ReplicableArchive), an OLTP system (CanonicalStore) and a separate read-only ReplicationServer which extracts updates from the database and provides them to a Replica used for OLAP. The OLTP part here is trivial and low-volume, and it's a proof-of-concept, so let's save ourselves a class by merging ReplicationServer into CanonicalStore.

LourensVeen:
This goes on top of #39, so that one needs to be merged first.

Closes #32.


svenvanderburg (approved):
Nice! This makes a lot of sense!

PR #46, opened 2020-11-23
Issue 18 intrinsic asset ids

LourensVeen:
This implements #18. It should be applied on top of #45, so ignore the first commit (or change the reference to `small-fixes`) because that's the other PR.

### DIFF
--- a/proof_of_concept/definitions/asset_id.py
+++ b/proof_of_concept/definitions/asset_id.py
@@ -3,7 +3,33 @@ from typing import Any, cast, Type
 
 
 class AssetId(str):
-    """An ID of an asset."""
+    """An id of an asset.
+
+    An Asset id is a string of any of the following forms:
+
+    1. id:<namespace>:<name>:<site>
+    2. id:<namespace>:<name>
+    3. hash:<hash>
+
+    Form 1 is used for data assets and compute assets, i.e. data sets
+    and software wrapped in container images. These are primary assets,
+    meaning they are put into the DDM from the outside, and they are
+    concrete assets, meaning there's an actual object (the container
+    image) associated with them, which is stored at <site>.
+
+    Form 2 is used for asset collections. These are named (with an
+    AssetId) collections of assets created implicitly by
+    InAssetCollection rules. These are also considered primary assets,
+    but they are abstract, not concrete, because there is no
+    downloadable object associated with them.
+
+    Form 3 is used for results of data processing done in the DDM. The
+    hash identifying the object is derived from the workflow used to
+    create this asset. These objects are called secondary assets, as
+    they are temporary, only existing during workflow execution. Their
+    id does not specify their location.
+
+    """
     def __new__(cls: Type['AssetId'], seq: Any) -> 'AssetId':
         """Create an AssetId.
 
--- a/proof_of_concept/definitions/asset_id.py
+++ b/proof_of_concept/definitions/asset_id.py
@@ -1,0 +1,60 @@
+"""Identifier for assets."""
+from typing import Any, cast, Type
+
+
+class AssetId(str):
+    """An ID of an asset."""
+    def __new__(cls: Type['AssetId'], seq: Any) -> 'AssetId':
+        """Create an AssetId.
+
+        Args:
+            seq: Contents, will be converted to a string using str(),
+            then used as the id.
+
+        Raises:
+            ValueError: If str(seq) is not a valid ID.
+        """
+        data = str(seq)
+        if data.startswith('id:'):
+            parts = len(data.split(':'))
+            if parts < 3 or parts > 4:
+                raise ValueError(f'Invalid asset id: {data}')
+        elif data.startswith('hash:'):
+            if not all([c in '0123456789abcdef' for c in data[5:]]):
+                raise ValueError(f'Invalid asset id: {data}')
+        elif data == '*':
+            # Valid as a wildcard in rules
+            pass
+        else:
+            raise ValueError(f'Invalid asset id type: {data}')
+
+        return str.__new__(cls, seq)        # type: ignore
+
+    @staticmethod
+    def from_key(key: str) -> 'AssetId':
+        """Creates an AssetId from a key (hash).
+
+        Args:
+            key: A hash of a workflow that created this asset.
+
+        Returns:
+            The AssetId for the workflow result.
+        """
+        return AssetId(f'hash:{key}')
+
+    def is_primary(self) -> bool:
+        """Returns whether this is a primary asset."""
+        return self.startswith('id:')
+
+    def namespace(self) -> str:
+        """Returns the namespace this asset is in.
+
+        Returns:
+            The namespace.
+
+        Raises:
+            RuntimeError: If this is not a primary asset.
+        """
+        if not self.is_primary():
+            raise RuntimeError('Namespace of secondary asset requested')
+        return self.split(':')[1]
--- a/proof_of_concept/definitions/asset_id.py
+++ b/proof_of_concept/definitions/asset_id.py
@@ -46,6 +46,14 @@ class AssetId(str):
         """Returns whether this is a primary asset."""
         return self.startswith('id:')
 
+    def is_concrete(self) -> bool:
+        """Returns whether this is a concrete asset.
+
+        An asset is concrete if it is primary and has a site it can
+        be downloaded from.
+        """
+        return self.is_primary() and len(self.split(':')) == 4
+
     def namespace(self) -> str:
         """Returns the namespace this asset is in.
 
@@ -58,3 +66,17 @@ class AssetId(str):
         if not self.is_primary():
             raise RuntimeError('Namespace of secondary asset requested')
         return self.split(':')[1]
+
+    def location(self) -> str:
+        """Returns the name of the site storing this asset.
+
+        Returns:
+            A site name.
+
+        Raises:
+            RuntimeError: If this is not a concrete asset.
+        """
+        if not self.is_concrete():
+            raise RuntimeError(
+                    'Location requested of non-concrete asset {self.data}')
+        return self.split(':')[3]
--- a/proof_of_concept/definitions/asset_id.py
+++ b/proof_of_concept/definitions/asset_id.py
@@ -42,8 +42,8 @@ class AssetId(str):
         """
         data = str(seq)
         if data.startswith('id:'):
-            parts = len(data.split(':'))
-            if parts < 3 or parts > 4:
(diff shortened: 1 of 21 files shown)


proof_of_concept/definitions/workflows.py
```diff
@@ -236,16 +244,19 @@ def calc_step_outputs(
...
-            step_hash.update(step.compute_asset_id.encode('utf-8'))
+                step_hash.update(
+                        item_keys[inp_item].encode('utf-8'))
+            step_hash.update(
+                    step.compute_asset_id.encode('utf-8'))
             for outp_name in step.outputs:
                 outp_item = '{}.{}'.format(step.name, outp_name)
                 outp_hash = step_hash.copy()
                 outp_hash.update(outp_name.encode('utf-8'))
-                item_keys[outp_item] = 'hash:{}'.format(outp_hash.hexdigest())
```
LourensVeen, line 245: Implementing the AssetId class revealed that `item_keys` had mostly keys in it, except for the step outputs, where it had asset ids. So I changed it to only have keys everywhere, and those keys get converted to AssetIds when needed (see the `from_key` calls above).

proof_of_concept/policy/evaluation.py
```diff
@@ -222,24 +223,12 @@ def set_input_assets_permissions(
 
             This modifies the permissions argument.
             """
-            for inp_asset in job.inputs.values():
+            for inp_name, inp_asset in job.inputs.items():
                 if inp_asset not in permissions:
```
LourensVeen: This is superfluous, fixed in later commit.

proof_of_concept/policy/evaluation.py
```diff
@@ -222,24 +223,12 @@ def set_input_assets_permissions(
...
-                    permissions[inp_asset] = (
+                    permissions[inp_name] = (
                             self._policy_evaluator.permissions_for_asset(
                                 inp_asset))
 
-        def prop_workflow_inputs(
```
LourensVeen, line 231: Removed because it was a bit ugly, the permissions are indexed by workflow item (e.g. workflow input, step input, step output, etc.) except that I also put in the job input assets, then propagated those to the inputs. Saves a bit of computation if you use the same input asset for multiple workflow inputs, but also makes the code less clean. So now I just look up the permissions for each workflow input and put them in directly, so that everything is indexed by workflow item and it all makes sense. Slightly less efficient but at this point that's not important (and will probably never be).

proof_of_concept/rest/serialization.py
```diff
@@ -147,18 +147,21 @@ def _deserialize_rule(user_input: JSON) -> Rule:
     """Deserialize a Rule from JSON."""
```
LourensVeen: Eh, all changes to this file are debris from editing. I've added a later commit which undoes the lot. Go ahead and skip this file.

proof_of_concept/definitions/asset_id.py
```diff
@@ -0,0 +1,82 @@
...
+        elif data == '*':
+            # Valid as a wildcard in rules
+            pass
+        else:
+            raise ValueError(f'Invalid asset id type: {data}')
+
+        return str.__new__(cls, seq)        # type: ignore
+
+    @staticmethod
+    def from_key(key: str) -> 'AssetId':
```
svenvanderburg: Shouldn't this be a classmethod?

proof_of_concept/definitions/asset_id.py
```diff
@@ -0,0 +1,82 @@
...
+            if parts < 3 or parts > 4:
+                raise ValueError(f'Invalid asset id: {data}')
+        elif data.startswith('hash:'):
+            if not all([c in '0123456789abcdef' for c in data[5:]]):
+                raise ValueError(f'Invalid asset id: {data}')
+        elif data == '*':
+            # Valid as a wildcard in rules
+            pass
+        else:
+            raise ValueError(f'Invalid asset id type: {data}')
```
svenvanderburg, line 55: Is this error message different on purpose?

proof_of_concept/definitions/asset_id.py
```diff
@@ -0,0 +1,82 @@
+"""Identifier for assets."""
+from typing import Any, cast, Type
+
+
+class AssetId(str):
+    """An ID of an asset."""
+    def __new__(cls: Type['AssetId'], seq: Any) -> 'AssetId':
+        """Create an AssetId.
```
svenvanderburg, line 34: It would be nice to add an example of a primary AssetId and a secondary AssetId, so it's easier to make sense of the code in here.

proof_of_concept/definitions/asset_id.py
```diff
@@ -0,0 +1,82 @@
...
+            then used as the id.
+
+        Raises:
+            ValueError: If str(seq) is not a valid ID.
+        """
+        data = str(seq)
+        if data.startswith('id:'):
+            parts = len(data.split(':'))
+            if parts < 3 or parts > 4:
+                raise ValueError(f'Invalid asset id: {data}')
```
svenvanderburg:
```suggestion
            n_parts = len(data.split(':'))
            if n_parts not in [3, 4]:
                raise ValueError(f'Invalid asset id: {data}')
```
Maybe this is clearer?

proof_of_concept/definitions/asset_id.py
```diff
@@ -0,0 +1,82 @@
...
+        """Returns whether this is a primary asset."""
+        return self.startswith('id:')
+
+    def is_concrete(self) -> bool:
+        """Returns whether this is a concrete asset.
+
+        An asset is concrete if it is primary and has a site it can
+        be downloaded from.
+        """
+        return self.is_primary() and len(self.split(':')) == 4
```
svenvanderburg, line 81: Could we think of a better name than `is_concrete`? `has_location` maybe? Something that tells us that it has a location.

proof_of_concept/definitions/asset_id.py
```diff
@@ -0,0 +1,82 @@
+"""Identifier for assets."""
+from typing import Any, cast, Type
+
+
+class AssetId(str):
```
svenvanderburg, line 5: Nice, I like how you implemented this :)

proof_of_concept/definitions/workflows.py
svenvanderburg: Bueno!

proof_of_concept/policy/evaluation.py
svenvanderburg: Sounds sane!

proof_of_concept/policy/replication.py
```diff
@@ -27,16 +32,17 @@ def __init__(self, namespace: str, key: RSAPublicKey) -> None:
...
         elif isinstance(rule, InAssetCollection):
-            namespace = rule.asset[3:].split('.')[0]
+            namespace = rule.asset.split(':')[1]
         elif isinstance(rule, InPartyCollection):
-            namespace = rule.collection[3:].split('.')[0]
+            namespace = rule.collection.split(':')[1]
 
+        logger.info(f'Rule namespace {namespace}, should be {self._namespace}')
         if namespace != self._namespace:
             return False
```
svenvanderburg, line 47: Can we somehow make use of the `AssetId.namespace` method here? Something like this (but then for all if/elif statements ofcourse)
```suggestion
            asset_id = AssetId(rule.collection)

        logger.info(f'Rule namespace {namespace}, should be {self._namespace}')
        if asset_id.namespace != self._namespace:
            return False
```

svenvanderburg (commented):
Great! This really makes things simpler and more scalable!

I left a few comments, but most looks really good!

proof_of_concept/definitions/asset_id.py
LourensVeen: Yes, an asset id needs to start with a tag that says which type of id it is, i.e. `id:` or `hash:`. In this branch, the type tag is wrong, which is what this error message tries to convey. I have a better error handling issue open somewhere, so we'll go and make this even better in the future if there's time.

proof_of_concept/definitions/asset_id.py
LourensVeen: I like this, more Pythonic and less C++ish. I'd prefer a tuple over a list though, `(3, 4)`. I cannot edit this it seems, but maybe you can?

proof_of_concept/definitions/asset_id.py
LourensVeen: Well, secondary assets (intermediate and final workflow results) also have a location, but not in their id. It's a subdivision of primary assets, which can be actual data sets and software (concrete) or collections of such (abstract). So maybe `has_location` is also misleading. Hmm.

proof_of_concept/policy/replication.py
LourensVeen: Actually, I have a note about adding a `namespace()` method to Rule, which would clean this up even more nicely, but I decided not to pile that on here.

proof_of_concept/definitions/asset_id.py
LourensVeen: Good point! C++-ism :-). Fixed in fcd6017

proof_of_concept/definitions/asset_id.py
LourensVeen: Fixed by  51c2767

proof_of_concept/definitions/asset_id.py
LourensVeen: Actually, getting subclassing str to work with mypy is still a bit of a mystery. I'm getting an error message that I shouldn't be getting I think, so I'm going to have to see if this is a known bug... but the `#type: ignore` works for now.

proof_of_concept/policy/replication.py
svenvanderburg: Bueno

proof_of_concept/definitions/asset_id.py
svenvanderburg: Check.

proof_of_concept/definitions/asset_id.py
svenvanderburg: Fair enough. If we cannot think of something better than `is_concrete` is the best name!

proof_of_concept/definitions/asset_id.py
svenvanderburg: Cool, yeah a tuple is nicer!

proof_of_concept/definitions/asset_id.py
svenvanderburg: Ok, nice!

PR #47, opened 2020-11-27
Add a RegistryServer class to the main code

### LINKED ISSUE #34: Add RegistryServer class to main code
This is now only in the tests, but should be in the main code for users who need a simple HTTP/REST server for the Registry as we have with SiteServer for sites.

LourensVeen:
This adds a RegistryServer class, using code from a test fixture. With this you can set up a Registry webservice from a script that sets up an experiment, if you just want to run something simple locally rather than having containers and multiple machines. Fixes #34.


svenvanderburg (approved):
Nice changes!

PR #48, opened 2020-11-27
Task 1244: Rule.signing_namespace()

LourensVeen:
This adds a `signing_namespace()` method to the Rule ABC, and implements it for the various rules. That simplifies the implementation of `RuleValidator`.

Also, we're now using an `AssetId` to identify a party collection, which sounds a bit odd. Maybe we should rename `AssetId` to `Identifier` and use it for parties and party collections as well; I've made a new PBI 1254 for this.

This applies on top of #47, so that should be merged first.

### DIFF
--- a/proof_of_concept/definitions/policy.py
+++ b/proof_of_concept/definitions/policy.py
@@ -4,4 +4,6 @@ from proof_of_concept.definitions.signable import Signable
 
 class Rule(Signable):
     """Abstract base class for policy rules."""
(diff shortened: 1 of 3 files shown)


svenvanderburg (approved):
This is really clean! 👍 

One comment: The changes in this PR don't really depend on changes in #47 even though you branched off from there. I think it would be easier to just branch off from 'develop'.

PR #49, opened 2020-11-27
Use __name__ to name loggers, not __file__

LourensVeen:
Not sure how we missed this, but some of the loggers are named using `__file__`, while the Python convention is to use `__name__` so you can control log output per module. This PR fixes that.

This goes on top of #48; ignore all but the top commit.


svenvanderburg (approved):
Looks good 👍  (but see my comment about branching in #48, it also applies to this branch)

PR #50, opened 2020-12-01
Clean up id/name/key/item terminology

LourensVeen:
This implements task 1245. It adds some definitions to the documentation, and then changes the code to match. The `AssetId` class has been renamed to `Identifier` and extended to cover identities of other kinds of objects. Most of this is just dumb renaming, so don't spend too much time on it. b1b28ce is the more interesting commit.

### DIFF
--- a/docs/terminology.rst
+++ b/docs/terminology.rst
@@ -2,6 +2,102 @@
 Terminology
 ===========
 
+Naming objects in and around the DDM
+====================================
+
+There are different kinds of objects in the DDM which need to be distinguished
+from other objects of the same kind. So they need some kind of unique
+identifier.
+
+Identifiers in the DDM are strings consisting of colon-separated components. The
+first component is a fixed string designating the type of object being
+identified. One or more components follow depending on the type (see below).
+Components may contain lower- and uppercase letters, digits, underscores,
+dashes, and periods.
+
+Here are some of those kinds of objects and how they are identified.
+
+Party
+  A party is a real-world entity which takes part in the DDM, usually an
+  organization. Parties may own or administer sites. Parties are uniquely
+  identified using an identifier of the form::
+
+    party:<namespace>:<name>
+
+  The ``<namespace>`` component is expected to be a DNS (sub-)domain, while
+  ``<name>`` can be any valid component.
+
+Party Collection
+  A party collection is a group of parties. It is used to organize parties and
+  make it easier to define rules that apply to multiple parties. Party
+  collections are identified using an identifier of the form::
+
+    party_collection:<namespace>:<name>
+
+  The ``<namespace>`` component is expected to be a DNS (sub-)domain, while
+  ``<name>`` can be any valid component.
+
+Site
+  A site is a running software installation which takes part in the DDM. Sites
+  have an owner (a Party) and an administrator (also a Party). Sites are
+  uniquely identified using an identifier of the form::
+
+    site:<namespace>:<name>
+
+  The ``<namespace>`` component is expected to be a DNS (sub-)domain, while
+  ``<name>`` can be any valid component.
+
+Asset
+  An asset is either a data set or a software component. Assets are identified
+  using an Identifier of the form::
+
+    asset:<namespace>:<name>:<site_namespace>:<site_name>
+
+  The ``<namespace>`` and ``<name>`` components are as above.
+  ``<site_namespace>`` and ``<site_name>`` refer to the site from which the
+  asset is available, ``site:<site_namespace>:<site_name>``.
+
+AssetCollection
+  An asset collection is a group of assets. It is used to organize assets and
+  make it easier to define rules that apply to multiple assets. Asset
+  collections are identified using an identifier of the form::
+
+    asset_collection:<namespace>:<name>
+
+  The ``<namespace>`` component is expected to be a DNS (sub-)domain, while
+  ``<name>`` can be any valid component.
+
+Result
+  A result is a data set that is an intermediate or final result of processing
+  data using a workflow. Results are identified by an identifier of the form::
+
+    result:<id_hash>
+
+  The ``<id_hash>`` is a string of letters and digits calculated from (the part
+  of) the workflow that was used to produce the result.
+
+Workflow item
+  A workflow item is a workflow input, workflow output, workflow step, workflow
+  step input, or workflow step output. In the context of dealing with
+  permissions for and execution of workflows, it's needed to refer to these.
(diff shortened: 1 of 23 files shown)


proof_of_concept/definitions/identifier.py
```diff
@@ -0,0 +1,101 @@
...
+    An Identifier is a string of any of the following forms:
+
+    1. party:<namespace>:<name>
+    2. party_collection:<namespace>:<name>
+    3. site:<namespace>:<name>
+    4. asset:<namespace>:<name>:<site_namespace>:<site_name>
+    5. asset_collection:<namespace>:<name>
+    6. result:<id_hash>
+
+    This class also accepts a single asterisk an an identifier, as it
```
svenvanderburg:
```suggestion
    This class also accepts a single asterisk as an identifier, as it
```

proof_of_concept/definitions/identifier.py
```diff
@@ -0,0 +1,101 @@
...
+    is used as a wildcard in rules.
+
+    See the Terminology section of the documentation for details.
```
svenvanderburg: Nice documentation! This is the minimum that you expect here, and in those cases you need more info you can read the Terminology documentation, awesome!

proof_of_concept/definitions/identifier.py
```diff
@@ -0,0 +1,101 @@
...
+        """
+        data = str(seq)
+
+        if data != '*':
+            parts = data.split(':')
+            if parts[0] not in cls._kinds:
+                raise ValueError(f'Invalid identifier kind {parts[0]}')
+
+            if len(parts) != cls._lengths[parts[0]]:
+                raise ValueError(f'Too few or too many parts in {data}')
```
svenvanderburg:
```suggestion
            parts = data.split(':')
            identifier_kind = parts[0]
            if identifier_kind not in cls._kinds:
                raise ValueError(f'Invalid identifier kind {parts[0]}')

            if len(parts) != cls._lengths[identifier_kind]:
                raise ValueError(f'Too few or too many parts in {data}')
```

svenvanderburg: To make this a bit less cryptic.

proof_of_concept/definitions/identifier.py
```diff
@@ -0,0 +1,101 @@
...
+            seq: Contents, will be converted to a string using str(),
+            then used as the identifier.
+
+        Raises:
+            ValueError: If str(seq) is not a valid identifier.
+        """
+        data = str(seq)
+
+        if data != '*':
+            parts = data.split(':')
```
svenvanderburg: There's the `parts` method for this right?

proof_of_concept/definitions/identifier.py
```diff
@@ -0,0 +1,101 @@
...
+
+        Raises:
+            RuntimeError: If this is not a primary asset.
+        """
+        if self.parts[0] == 'result':
+            raise RuntimeError('Results do not have a namespace')
+        return self.parts[1]
+
+    def location(self) -> 'Identifier':
+        """Returns the d of the site storing this asset.
```
svenvanderburg: the d?

proof_of_concept/definitions/identifier.py
```diff
@@ -0,0 +1,101 @@
+"""Identifier for various things in the DDM."""
+import re
+from typing import Any, cast, List, Type
+
+
+class Identifier(str):
```
svenvanderburg, line 6: Maybe you can refer to `components` instead of `parts` to match the documentation of terminology.

docs/terminology.rst
```diff
@@ -2,6 +2,102 @@
 Terminology
 ===========
 
+Naming objects in and around the DDM
```
svenvanderburg, line 5: This documentation *could* be less verbose, using for example more lists of keywords instead of full sentences. But I already like it a lot the way it is, so you can also just leave it as is.

svenvanderburg (approved):
Nice changes! I've left a few minor comments.

proof_of_concept/definitions/identifier.py
LourensVeen: Yes, but we can't call a method on an object before it's been constructed...

proof_of_concept/definitions/identifier.py
LourensVeen: Apparently every asset has a d, which is being returned here... :-) No idea what happened there, but it's fixed.

proof_of_concept/definitions/identifier.py
LourensVeen: Yeah. The issue is that "component" is already a package, and I don't like overloading terms. I tried changing the documentation instead, replacing "component" with "part" there, but that reads a bit weird. I'm thinking to try "section" now.

docs/terminology.rst
LourensVeen: I disagree, I think making it more terse would make it harder to read. The vocabulary for ODRL has a few nice examples of terse definitions that are difficult to comprehend because there's no context, e.g. https://www.w3.org/TR/odrl-vocab/#term-informingParty. And the parts of it that are understandable are at least as verbose, e.g. https://www.w3.org/TR/odrl-model/#asset.

proof_of_concept/definitions/identifier.py
svenvanderburg: Never mind...

proof_of_concept/definitions/identifier.py
svenvanderburg: lol

proof_of_concept/definitions/identifier.py
svenvanderburg: Fair enough. Maybe `splits` or `elements`? Or `section` is also fine I guess.

LourensVeen: Pieces, parts, fragments, shards, sections or components.
A segment or a unit or an item?
The hardest task in programming
is not solved in Python.

LourensVeen: Segments it is, it kind of looks like a worm anyway :-).

proof_of_concept/definitions/identifier.py
LourensVeen: Sorry, can't be bothered. It's good enough as is.

proof_of_concept/definitions/identifier.py
svenvanderburg: Lol! Are you in sinterklaas mode?

PR #51, opened 2020-12-03
Add a WSGI app function for the registry

LourensVeen:
This adds a function which returns a WSGI app for a WSGI server such as gunicorn to use. Completes task 1247.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -2,7 +2,7 @@
 # -*- coding: utf-8 -*-
 import os
 
-from setuptools import setup
+from setuptools import find_packages, setup
 
 here = os.path.abspath(os.path.dirname(__file__))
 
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -43,7 +41,8 @@
     install_requires=[
         'cryptography',
         'python-dateutil',
-        'falcon',
+        'requests',
+        'falcon==3.0.0a3',
```
bakhshir, line 45: Why did you need to fix the version of falcon?

bakhshir (approved):
All the changes look reasonable. Since I'm reviewing for the first time: should i run the installation procedure to validate? ;-)

setup.py
LourensVeen: The reason for the fixed version is that 2.0.0 is the current stable version, but there's a 3.0.0a3 alpha release. If you just `pip install` then you get version 2, because `pip` understands alpha versions and by default doesn't install them. If you run `setup.py test`, it will use `easy_install` to build a test environment, and `easy_install` is ancient and doesn't understand alpha versions, so you get version 3. And that has a different API. Which is actually what I accidentally programmed against apparently, so everything worked fine when running the tests, but then when I started building the Docker I used pip and it didn't work, because that installs version 2...

I'm considering whether to backport to version 2, or to just wait for 3 to get released, but in the mean time, this works at least.

PR #52, opened 2020-12-03
Add registry docker image build

LourensVeen:
This adds a Makefile and some scripts which can be used to create a Docker image which runs the registry component as a REST service. Merge #51 as well, because this doesn't work without that.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,10 +1,10 @@
-.PHONY: docker
-docker: registry_docker
+.PHONY: docker_images
+docker_images: registry_docker_image
 
 .PHONY: docker_clean
 docker_clean:
-	docker rmi mahiru-registry:latest
+	docker rmi -f mahiru-registry:latest
 
-.PHONY: registry_docker
-registry_docker:
+.PHONY: registry_docker_image
+registry_docker_image:
 	docker build . -f docker/registry.Dockerfile -t mahiru-registry:latest
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,10 @@
+.PHONY: docker
+docker: registry_docker
+
+.PHONY: docker_clean
+docker_clean:
+	docker rmi mahiru-registry:latest
+
+.PHONY: registry_docker
+registry_docker:
+	docker build . -f docker/registry.Dockerfile -t mahiru-registry:latest
--- a/docker/init.sh
+++ b/docker/init.sh
@@ -8,6 +8,8 @@ function stop_container {
     kill -TERM $(cat ${pid_file})
 }
 
+# Docker sends us a SIGTERM on 'docker stop'.
+# This catches it and shuts down cleanly by calling the function above
 trap stop_container SIGTERM
 
 echo 'Installed handler' 1>&2
--- a/docker/init.sh
+++ b/docker/init.sh
@@ -1,0 +1,21 @@
+#!/bin/bash
+
+pid_file='/var/run/gunicorn/gunicorn.pid'
+
+function stop_container {
+    service nginx stop
+    gunicorn_pid=$(cat ${pid_file})
+    kill -TERM $(cat ${pid_file})
+}
+
+trap stop_container SIGTERM
+
+echo 'Installed handler' 1>&2
+
+service nginx start
+
+su -c ". /home/mahiru/.env/bin/activate && gunicorn --pid ${pid_file} --access-logfile /var/log/gunicorn/access.log --error-logfile /var/log/gunicorn/error.log --capture-output --bind 127.0.0.1:8000 'proof_of_concept.rest.registry:wsgi_app()'" mahiru &
+
+wait
+
+rm -f ${pid_file}
--- a/docker/init.sh
+++ b/docker/init.sh
@@ -14,7 +14,7 @@ echo 'Installed handler' 1>&2
 
 service nginx start
 
-su -c ". /home/mahiru/.env/bin/activate && gunicorn --pid ${pid_file} --access-logfile /var/log/gunicorn/access.log --error-logfile /var/log/gunicorn/error.log --capture-output --bind 127.0.0.1:8000 'proof_of_concept.rest.registry:wsgi_app()'" mahiru &
+su -c "/home/mahiru/.local/bin/gunicorn --pid ${pid_file} --access-logfile /var/log/gunicorn/access.log --error-logfile /var/log/gunicorn/error.log --capture-output --bind 127.0.0.1:8000 'proof_of_concept.rest.registry:wsgi_app()'" mahiru &
 
 wait
 
--- a/docker/init.sh
+++ b/docker/init.sh
@@ -1,5 +1,6 @@
 #!/bin/bash
 
+wsgi_entry_point="$1"
 pid_file='/var/run/gunicorn/gunicorn.pid'
 
 function stop_container {
@@ -16,7 +17,7 @@ echo 'Installed handler' 1>&2
 
 service nginx start
 
-su -c "/home/mahiru/.local/bin/gunicorn --pid ${pid_file} --access-logfile /var/log/gunicorn/access.log --error-logfile /var/log/gunicorn/error.log --capture-output --bind 127.0.0.1:8000 'proof_of_concept.rest.registry:wsgi_app()'" mahiru &
+su -c "/home/mahiru/.local/bin/gunicorn --pid ${pid_file} --access-logfile /var/log/gunicorn/access.log --error-logfile /var/log/gunicorn/error.log --capture-output --bind 127.0.0.1:8000 '${wsgi_entry_point}'" mahiru &
 
 wait
 
--- a/docker/init.sh
+++ b/docker/init.sh
@@ -17,7 +17,7 @@ echo 'Installed handler' 1>&2
 
 service nginx start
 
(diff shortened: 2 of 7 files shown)


docker/registry.Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+COPY docker/nginx.conf /etc/nginx/sites-available/default
+COPY docker/init.sh /usr/local/bin/init.sh
+RUN chmod +x /usr/local/bin/init.sh
+
+COPY . /home/mahiru/
+RUN chown -R mahiru:mahiru /home/mahiru
+USER mahiru
+RUN \
+    python3 -m venv /home/mahiru/.env && \
+    . /home/mahiru/.env/bin/activate && \
```
svenvanderburg: Do you need a virtualenv? You can just install everything on sytem python right?

docker/registry.Dockerfile
```diff
@@ -0,0 +1,31 @@
+FROM ubuntu:20.04
```
svenvanderburg, line 1: Maybe you can add some comments why you need certain things. I bet in half a year you don't remember:
* Why you need this user mahiru and a home folder
* Why you need the `/var/log` and `/var/run` folders

docker/nginx.conf
```diff
@@ -0,0 +1,7 @@
+server {
```
svenvanderburg, line 1: Is there anything about this that you should document? (In a small comment)

docker/init.sh
```diff
@@ -0,0 +1,21 @@
...
+
+pid_file='/var/run/gunicorn/gunicorn.pid'
+
+function stop_container {
+    service nginx stop
+    gunicorn_pid=$(cat ${pid_file})
+    kill -TERM $(cat ${pid_file})
+}
+
+trap stop_container SIGTERM
```
svenvanderburg, line 14: Maybe document why you need to do this?

docker/init.sh
```diff
@@ -0,0 +1,21 @@
...
+
+echo 'Installed handler' 1>&2
+
+service nginx start
+
+su -c ". /home/mahiru/.env/bin/activate && gunicorn --pid ${pid_file} --access-logfile /var/log/gunicorn/access.log --error-logfile /var/log/gunicorn/error.log --capture-output --bind 127.0.0.1:8000 'proof_of_concept.rest.registry:wsgi_app()'" mahiru &
```
svenvanderburg: All the arguments here could be arguments for this script, which you could set in the docker file.

svenvanderburg: Why do you have to activate the venv in this script? That is done by the docker image right?

Makefile
```diff
@@ -0,0 +1,10 @@
+.PHONY: docker
```
svenvanderburg: Maybe you could call this `build` (and `registry_docker` `build_registry_docker`). Then also add a `run` that runs the docker image.

svenvanderburg (approved):
Cool! Dockerizing is an important next step! I left some comments, mostly about clarification and documentation.

docker/registry.Dockerfile
LourensVeen: Well, running services as root is generally bad practice, so you see this making a user in every Dockerfile. I guess I could have gone with `/opt/mahiru` instead of `/home/mahiru` but it doesn't really matter. Logging is very useful for debugging and `/var/log` is where logs live according to the [FHS](https://refspecs.linuxfoundation.org/FHS_3.0/fhs/index.html), and `/var/run` is where you put your pid files according to same. I looked at how I [did this for Cerise](https://github.com/MD-Studio/cerise/blob/develop/conf/docker-init.sh) two years ago and still remembered how and why, so I think I'll be okay in six months.

I think anyone familiar with Docker files and basic GNU/Linux system administration will be able to figure this one out, it's nothing special.

docker/nginx.conf
LourensVeen: I don't know, is there? To me it pretty much speaks for itself: we listen on port 80, then act as a proxy and pass on requests for `/` to whatever's listening on port 8000. Maybe there could be a [link to the nginx docs](http://nginx.org/en/docs/beginners_guide.html), but it's the first result in Google if I search for `nginx configuration`, so...

docker/init.sh
LourensVeen: Yeah, I guess this deserves a comment; most people probably just let their container get killed.

docker/init.sh
LourensVeen: Ah, good point. Most of them are fine as-is, but I had noticed that the app function entry point is the only difference between the registry image and the site image. So moving that to the Docker file would save some duplication. Will fix.

LourensVeen: Activating a venv changes environment variables. Those are not stored in a Docker layer, so you have to activate it here again.

Makefile
LourensVeen: But that's not good make terminology. You make a target, so a target should be a noun, not a verb. I think I'll change it to `registry_docker_image` though. As for running a DDM based on the images, that will take a bit more work because we need a way to load assets and submit requests and such. That's for some future PR.

docker/registry.Dockerfile
LourensVeen: I tried, but had some issues. I'll give it another go because it is a better solution, and it's always easier to start from working code :-).

docker/registry.Dockerfile
LourensVeen: Maybe this is something we need to discuss with the team actually, but I like this quote from Google's Python style guide: "On the other hand, never describe the code. Assume the person reading the code knows Python (though not what you’re trying to do) better than you do." Writing tutorials are fine, but it should be kept separate from programming.

LourensVeen:
Note: this was originally an independent branch off of `develop`, but since it depends on #51 that caused it to not work (it passed the CI because the CI doesn't test the Docker image). As a result, I couldn't test the changes suggested in the review. So I rebased it on top of #51, so that now there are two additional commits from that one. Conclusion: making independent branches doesn't work, stacking is the way to go. I'm going to try to do merges locally on the command line rather than using the GUI, see if that helps avoiding a mess when the reference gets changed.

Makefile
svenvanderburg: OK, I guess I have always abused `make` for my evil data science purposes :)

What I mean with `run` is to run this docker image to serve the registry, that should work right?

docker/nginx.conf
svenvanderburg: Check 👍

docker/registry.Dockerfile
svenvanderburg, line 1: You could also think of using something like: https://github.com/tiangolo/uwsgi-nginx-docker

svenvanderburg: I totally agree to never describe the code. But *why* you need code is different from that.

I don't think you can assume that all future developers will know all best practices for docker. I never created users in my docker images nor home folders, probably because it makes things overly complicated and I am lazy. That's why I was wondering whether there was a particular reason for it.

But feel free to leave it like this!

docker/registry.Dockerfile
LourensVeen: Yay, managed to fix it :-).

docker/registry.Dockerfile
LourensVeen: I don't think the point of that statement is to point out that everyone at Google is an expert at Python. There has to be *someone* at Google who is their least-experienced Python developer, and who may therefore not be able to immediately understand the code their colleagues are writing.

The point is that if you require the programmer to document everything that may be unknown to someone, they'll end up starting every docstring with "Python is a programming language. Programming languages are used to ...". There's no reasonable limit, and you end up drowning the code in explanations that make the code harder to read for more experienced developers, all to support a beginner programmer who may never actually exist.

Clearly, if a less experienced programmer needs to work on the code, then they may run into language features or idiom or algorithms that they don't understand. And then, in order to be able to do their job, they're going to have to learn about this unknown thing. But they should do so by looking up a tutorial or a blog post or the documentation, and those should be separate objects. The purpose of code is not to teach people how to program (unless it was written for a tutorial), and trying to tack on that responsibility will make the code less clear.

So, as a specific example, if you saw this Dockerfile and you wondered why it didn't just run as root, then you could do a web search and find [this blog post](https://medium.com/@mccode/processes-in-containers-should-not-run-as-root-2feae3f0df3b) that explains things much better than any comment could.

Makefile
LourensVeen: Yeah, that could work, but the question is what you would use a bare registry for. I imagine there being some kind of driver script that starts up a registry and some sites and then controls the whole thing to run an experiment, but I'm not sure how that would work yet, and I think it's better to do in Python than using Make. So I'm thinking to just stick with building, and leave running for the future.

docker/init.sh
LourensVeen: Added one...

docker/registry.Dockerfile
LourensVeen: I guess I could, I've never used uWSGI. I've looked at the documentation a bit, but I can't find whether it uses processes or threads and how to configure that, other than that it seems to use two processes by default. Since the Python code contains the registry state it cannot run in a multiprocess model. (We'd need a real database backend to make that work, but I haven't implemented that.) So too much effort for now, but I'll keep it in mind.

svenvanderburg: I think these are all good points. I do think that just putting a small comment doesn't really do harm, and can often clarify a lot and prevent people for having to search context to understand code. But let's not discuss the universe on this PR and let's agree to 90% agree and 10% disagree ;)

svenvanderburg: Nice blog post by the way :)

PR #53, opened 2020-12-07
Add site WSGI app

LourensVeen:
This adds a WSGI app for a site, This doesn't have assets or policies, so it's only useful as a trusted third party, but it's a good start, more will come in the future. Starts with 1021c1c actually, I'll merge the previous branches tomorrow and then the older commits should disappear.

### DIFF
--- a/proof_of_concept/rest/ddm_site.py
+++ b/proof_of_concept/rest/ddm_site.py
@@ -7,7 +7,10 @@ from wsgiref.simple_server import WSGIRequestHandler, WSGIServer
 
 from falcon import App, HTTP_200, HTTP_400, HTTP_404, Request, Response
 import ruamel.yaml as yaml
+import yatiml
 
+from proof_of_concept.components.ddm_site import Site
+from proof_of_concept.components.registry_client import RegistryClient
 from proof_of_concept.definitions.identifier import Identifier
 from proof_of_concept.definitions.interfaces import IAssetStore, IStepRunner
 from proof_of_concept.definitions.policy import Rule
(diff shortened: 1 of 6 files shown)


proof_of_concept/rest/ddm_site.py
```diff
@@ -184,3 +187,44 @@ def close(self) -> None:
...
+        self.owner = owner
+        self.endpoint = endpoint
+
+
+load_settings = yatiml.load_function(Settings, Identifier)
+
+
+def wsgi_app() -> App:
+    """Creates a WSGI app for a WSGI runner."""
+    settings = load_settings(Path('/etc/mahiru/mahiru.conf'))
```
svenvanderburg, line 224: Maybe make the path a constant?

tests/test_scenarios.py
```diff
@@ -2,26 +2,129 @@
...
+                desc['rules'], registry_client)
+            for site_name, desc in site_descriptions.items()}
+
+
+def create_servers(sites: Dict[str, Site]) -> Dict[str, SiteServer]:
+    """Create REST servers for sites."""
+    return {
+            site_name: SiteServer(
+                    SiteRestApi(site.policy_store, site.store, site.runner))
+            for site_name, site in sites.items()}
```
svenvanderburg, line 72: If I understand it correctly this is now different code than what a user would use (then you use `wsgi_app` based on settings right?). And I believe I remember from the RegistryServer code that you somehow reused more of the code to start the server in the tests. Maybe you could try something like that here?

svenvanderburg (approved):
Nice changes! I left a few minor comments.

tests/test_scenarios.py
LourensVeen: Well, this test uses the prototype in everything-in-one-process mode, while the `wsgi_app` is for use in distributed mode. 

There will eventually be a similar driver for the distributed set-up, so there will be a docker-compose (or ansible or something, if you're running across multiple machines) there that starts up a registry and a couple of sites, and then some containers that represent users which connect to a site and submit assets and workflows and policies. But I need to build a few more REST APIs before that can work.

So yeah, maybe eventually this will get turned into an experiment definition support library for the local running case, but for now it just needs to run these tests.

svenvanderburg: Cool!

PR #54, opened 2020-12-07
Add docker image for a site

LourensVeen:
This adds a Docker image for a site (almost completely the same as the registry image, just a different endpoint) and a docker-compose file that starts a registry and a site, and has the site connect to the registry. It also adds auto-reconnect to replication sources, because without that you get in trouble when trying to docker-compose two containers together (and it's a good idea anyway, in a distributed system). I should probably use retrying for that. Maybe tomorrow.

Goes on top of #53, please don't merge as the target is set for reviewing, it's not where this will eventually go.

### DIFF
--- a/docker/base.Dockerfile
+++ b/docker/base.Dockerfile
@@ -1,0 +1,23 @@
+FROM ubuntu:20.04
+RUN \
+    apt-get update && \
+    apt-get -y upgrade && \
+    apt-get -y install nginx && \
+    apt-get -y install python3 python3-pip python3-venv python3-wheel
+RUN \
+    useradd mahiru && \
+    mkdir /home/mahiru && \
+    chown -R mahiru:mahiru /home/mahiru
+RUN \
+    mkdir /var/log/gunicorn && \
+    chown mahiru:mahiru /var/log/gunicorn && \
+    mkdir /var/run/gunicorn && \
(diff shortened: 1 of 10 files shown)


docker/base.Dockerfile
```diff
@@ -0,0 +1,23 @@
+FROM ubuntu:20.04
```
svenvanderburg, line 1: Nice to use a base 👍

docker/docker-compose.yml
```diff
@@ -0,0 +1,12 @@
+version: "3.3"
```
svenvanderburg, line 1: Yeah! A real working DDM!

docker/site.conf
```diff
@@ -0,0 +1,4 @@
+name: site
```
svenvanderburg, line 1: Maybe rename this file to `example-site.conf`?

svenvanderburg (approved):
Looks good! 👍  Just a minor comment

docker/site.conf
LourensVeen: Well, eventually I hope this grows into the real thing. Yeah, it's not very usable at the moment, but it's not intended to be an example.

LourensVeen:
No, it was merged alright, but into develop, which was the intent.

PR #55, opened 2021-01-25
Store images in asset store

LourensVeen:
This gives the AssetStore the ability to store Docker images, which I think covers task 1347.

I had a look at using the standard nginx Docker image as a base for the data images, but couldn't get it to work quickly. The current setup is also not exactly exotic, just Ubuntu + nginx with WebDAV enabled, which is a configuration I'm familiar with, so I decided to go with that for now. The whole point of a Docker container is of course that it encapsulates the stuff inside, so this can always be changed later if we're so inclined.

### DIFF
--- a/proof_of_concept/components/asset_store.py
+++ b/proof_of_concept/components/asset_store.py
@@ -1,6 +1,10 @@
 """Storage and exchange of data and compute assets."""
+from copy import copy
 import logging
-from typing import Dict
+from pathlib import Path
+from shutil import copyfile, move, rmtree
+from tempfile import mkdtemp
+from typing import Dict, Optional
 
 from proof_of_concept.definitions.assets import Asset
 from proof_of_concept.definitions.identifier import Identifier
@@ -14,17 +18,29 @@ logger = logging.getLogger(__name__)
 
 class AssetStore(IAssetStore):
     """A simple store for assets."""
-    def __init__(self, policy_evaluator: PolicyEvaluator) -> None:
+    def __init__(
+            self, policy_evaluator: PolicyEvaluator,
+            image_dir: Optional[Path] = None) -> None:
         """Create a new empty AssetStore."""
         self._policy_evaluator = policy_evaluator
         self._permission_calculator = PermissionCalculator(policy_evaluator)
         self._assets = dict()  # type: Dict[Identifier, Asset]
+        if image_dir is None:
+            image_dir = Path(mkdtemp())
+        self._image_dir = image_dir
 
-    def store(self, asset: Asset) -> None:
+    def close(self) -> None:
+        """Releases resources, call when done."""
+        rmtree(self._image_dir, ignore_errors=True)
+
+    def store(self, asset: Asset, move_image: bool = False) -> None:
         """Stores an asset.
 
         Args:
             asset: asset object to store
+            move_image: If the asset has an image and True is passed,
+                the image file will be moved rather than copied into
+                the store.
 
         Raises:
             KeyError: If there's already an asset with the asset id.
@@ -33,10 +49,17 @@ class AssetStore(IAssetStore):
         if asset.id in self._assets:
             raise KeyError(f'There is already an asset with id {id}')
 
-        self._assets[asset.id] = asset
+        self._assets[asset.id] = copy(asset)
+        if asset.image_location is not None:
+            src_path = Path(asset.image_location)
+            tgt_path = self._image_dir / f'{asset.id}.tar.gz'
+            if move_image:
+                move(str(src_path), str(tgt_path))
+            else:
+                copyfile(src_path, tgt_path)
+            self._assets[asset.id].image_location = str(tgt_path)
 
-    def retrieve(self, asset_id: Identifier, requester: str
-                 ) -> Asset:
+    def retrieve(self, asset_id: Identifier, requester: str) -> Asset:
         """Retrieves an asset.
 
         Args:
--- a/proof_of_concept/definitions/identifier.py
+++ b/proof_of_concept/definitions/identifier.py
@@ -44,7 +44,7 @@ class Identifier(str):
                 if not cls._segment_regex.match(segment):
                     raise ValueError(f'Invalid identifier segment {segment}')
 
-        return str.__new__(cls, seq)        # type: ignore
+        return str.__new__(cls, seq)
 
     @classmethod
     def from_id_hash(cls, id_hash: str) -> 'Identifier':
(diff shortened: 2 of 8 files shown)


LourensVeen:
Converted back to draft, I think I can make this more reviewable...

LourensVeen:
I tried to make it a bit smaller, it's still a fair amount of code but I think it could work. I guess a single image would be enough just to test storing images, but I have the rest now so we might as well I guess. Feel free to have a go at reviewing :).

proof_of_concept/components/asset_store.py
```diff
@@ -14,17 +18,29 @@
...
+        if image_dir is None:
+            image_dir = Path(mkdtemp())
+        self._image_dir = image_dir
 
-    def store(self, asset: Asset) -> None:
+    def close(self) -> None:
+        """Releases resources, call when done."""
+        rmtree(self._image_dir, ignore_errors=True)
+
+    def store(self, asset: Asset, move_image: bool = False) -> None:
```
bakhshir, line 36: AssetStore.store sounds a bit funny to me, but i can't suggest alternative ('put' does not sound good)

proof_of_concept/components/asset_store.py
```diff
@@ -33,10 +49,17 @@ def store(self, asset: Asset) -> None:
         if asset.id in self._assets:
             raise KeyError(f'There is already an asset with id {id}')
 
-        self._assets[asset.id] = asset
+        self._assets[asset.id] = copy(asset)
```
bakhshir, line 52: looking back and forth in the library (because i don't know it), this seems ok

proof_of_concept/components/asset_store.py
bakhshir, line 36: AssetStore.store sounds a bit funny, but i could not suggest any better alternative ('put' is also meh).

proof_of_concept/definitions/identifier.py
```diff
@@ -44,7 +44,7 @@ def __new__(cls: Type['Identifier'], seq: Any) -> 'Identifier':
                 if not cls._segment_regex.match(segment):
                     raise ValueError(f'Invalid identifier segment {segment}')
 
-        return str.__new__(cls, seq)        # type: ignore
+        return str.__new__(cls, seq)
```
bakhshir, line 47: does not need to be ignored? ;-)

bakhshir (commented):
Adding docker images handling seems fine to me, I'm not familiar with the dockerization and tests related to it, so I leave it to Sven to review. Otherwise, let me know, I can make an effort.

proof_of_concept/components/asset_store.py
LourensVeen: The idea here is that because the asset image is moved or copied, I have to update the Asset object to point to the new location. I don't want to mess up the Asset object passed by the caller, so I'm copying it before modifying. Previously that wasn't necessary, because the Asset wasn't altered.

proof_of_concept/components/asset_store.py
LourensVeen: Hmm, yes, English likes to verb its nouns... Maybe I could rename AssetStore to AssetWarehouse, AssetDepot, AssetBank, AssetDatabase or AssetMuseum (for curated data only!)? Although I guess for AssetDepot I'd have to go with AssetDepot.deposit(), so that may not solve it either :-).

proof_of_concept/definitions/identifier.py
LourensVeen: Ah yes, welcome to mypy. This is the correct way to create a class derived from str (I think, it's not very extensively documented), but the problem is that this statement returns an object of type `str`, while the signature wants an `Identifier`, so mypy 0.760 gave an error. I'm not sure actually why I used `# type ignore` instead of a cast, but that disabled the error.

Recently, mypy 0.800 was released, and apparently it now knows about deriving from `str` making the `# type: ignore` superfluous. And since superfluous `# type ignore` statements are an error in mypy, I removed it so the tests would pass...

tests/test_containers.py
```diff
@@ -0,0 +1,158 @@
...
+@pytest.fixture
+def dcli():
+    """Docker client object."""
+    return docker.from_env()
+
+
+@pytest.fixture
+def docker_dir(test_dir):
+    """Path to the tests/docker directory."""
+    return test_dir / 'docker'
```
svenvanderburg: I usually use constants for filepaths such as this. IMO it is more explicit and less verbose than fixtures. What is the advantage of using fixtures for this?

svenvanderburg (approved):
Nice progress! Using ubuntu + nginx + webdav sounds sane.

The changes to `assets.py` and `asset_store.py` are nice and clean. The 'testing' code is a bit less clean. I am not sure how permanent those changes are:
If sort-of permanent :the docker files could benefit from a short documentation comment and `tests/docker/app.py` could maybe be cleaner and clearer with some functions factored out and a bit more documentation.

tests/test_containers.py
LourensVeen: Mainly symmetry with pytest's `tmpdir` fixture I guess, and my general distaste for globals :-)

LourensVeen:
I've dockumented the docker files and while I was at it some of the stuff that goes in them, and I've refactored `app.py` into a `libmahiru.py` and `app.py` and cleaned it up a bit. Also, the new `libmahiru.py` now has a global constant _and_ a global variable! Better? :-)

LourensVeen:
@bakhshir Can I ask you to have a look at the new `test_asset_store.py`? I decided to remove the Docker-based tests (they'll be back later when I implement execution) and just have a unit test here, as that seems more appropriate.

tests/test_asset_store.py
```diff
@@ -0,0 +1,52 @@
+from pathlib import Path
+from unittest.mock import MagicMock
```
bakhshir, line 2: I was reading about MagicMock on friday :)

bakhshir (approved):
test_asset_store.py  looks good, and definitely better readable.

tests/test_asset_store.py
LourensVeen: It's aptly named, it really is magic. I'm still not entirely sure I understand it, but i can get it to work most of the time, and it's super handy...

bakhshir: I was reading on Friday Mock vs MagicMock. Very interesting, but indeed needs some practice to really get the feeling for it. I get to know quite a bit from the pr reviews, i like it! ;-)

bakhshir (approved):
sorry, pressed the wrong button :-)

PR #56, opened 2021-02-02
Add support for exchanging containers between sites

LourensVeen:
This adds a new endpoint to the site REST API that allows downloading a container image for a container. It's a separate endpoint (as opposed to serialising the container inside the Asset object) because the image is a large (hundreds of megabytes at least) binary blob. A separate endpoint avoids inefficient encodings and makes it easy to download the image straight to disk. Also, in the future we may want to use a network connection instead of downloading the image, so it needs to be a separate step.

### DIFF
--- a/proof_of_concept/definitions/assets.py
+++ b/proof_of_concept/definitions/assets.py
@@ -31,6 +31,8 @@ class Metadata:
 class Asset:
     """Asset, a representation of a computation or piece of data."""
 
+    KINDS = ('compute', 'data')
+
     def __init__(self, id: Union[str, Identifier], kind: str, data: Any,
                  image_location: Optional[str] = None,
                  metadata: Optional[Metadata] = None
@@ -54,7 +56,7 @@ class Asset:
         """
         if not isinstance(id, Identifier):
             id = Identifier(id)
-        if kind not in ('data', 'compute'):
+        if kind not in self.KINDS:
             raise ValueError(f'Invalid kind {kind}')
         if metadata is None:
             metadata = Metadata(Job.niljob(id), 'dataset')
--- a/proof_of_concept/definitions/assets.py
+++ b/proof_of_concept/definitions/assets.py
@@ -31,18 +31,19 @@ class Metadata:
 class Asset:
     """Asset, a representation of a computation or piece of data."""
 
-    def __init__(self, id: Union[str, Identifier], data: Any,
+    def __init__(self, id: Union[str, Identifier], kind: str, data: Any,
                  image_location: Optional[str] = None,
                  metadata: Optional[Metadata] = None
                  ) -> None:
         """Create an Asset.
 
         Assets may have either a data item (a JSON-serialisable Python
-        value) associated with them, or reference Docker image tarball
+        value) associated with them, or reference a Docker image tarball
         containing data or code.
 
         Args:
             id: Identifier of the asset
+            kind: Either "data" or "compute"
             data: Data related to the asset
             image_location: URL or path to the container image file.
             metadata: Metadata related to the asset. If no metadata is
@@ -53,9 +54,12 @@ class Asset:
         """
         if not isinstance(id, Identifier):
             id = Identifier(id)
+        if kind not in ('data', 'compute'):
+            raise ValueError(f'Invalid kind {kind}')
         if metadata is None:
             metadata = Metadata(Job.niljob(id), 'dataset')
         self.id = id
+        self.kind = kind
         self.data = data
         self.image_location = image_location
         self.metadata = metadata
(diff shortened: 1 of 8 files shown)


proof_of_concept/definitions/assets.py
```diff
@@ -53,16 +54,25 @@ def __init__(self, id: Union[str, Identifier], data: Any,
         """
         if not isinstance(id, Identifier):
             id = Identifier(id)
+        if kind not in ('data', 'compute'):
```
svenvanderburg: You could extract a constant `ASSET_KINDS`

proof_of_concept/definitions/assets.py
```diff
@@ -31,18 +31,19 @@ def __init__(self, job: Job, item: str) -> None:
...
         """Create an Asset.
 
         Assets may have either a data item (a JSON-serialisable Python
-        value) associated with them, or reference Docker image tarball
+        value) associated with them, or reference a Docker image tarball
         containing data or code.
 
         Args:
             id: Identifier of the asset
+            kind: Either "data" or "compute"
```
svenvanderburg, line 48: Could we not instead use the type of an object (i.e. `ComputeAsset` or `DataAsset`) to know the kind of asset? Or did you already try that and it didn't work out?

proof_of_concept/rest/client.py
```diff
@@ -52,6 +53,35 @@ def retrieve_asset(self, site_id: Identifier, asset_id: Identifier
...
+                stream=True) as r:
+            if r.status_code == 404:
+                raise KeyError('Asset image not found')
+            elif not r.ok:
+                raise RuntimeError('Server error when retrieving asset image')
+
+            if r.headers.get('Transfer-Encoding', '') == 'chunked':
+                chunk_size = None
+            else:
+                chunk_size = 1024 * 1024
```
svenvanderburg, line 78: You could extract a constant

proof_of_concept/definitions/assets.py
svenvanderburg: Ah I see we need this for deserialisation right?

tests/test_asset_exchange.py
```diff
@@ -0,0 +1,68 @@
...
+@pytest.fixture
+def asset_store(temp_path, test_image_file, asset_id):
+    asset_store = AssetStore(MagicMock(), temp_path)
+    asset = DataAsset(asset_id, None, str(test_image_file))
+    asset_store.store(asset)
+    return asset_store
+
+
+@pytest.fixture
+def image_server(asset_store):
```
svenvanderburg, line 34: Why do we need a separate image server here? The AssetImageAccessHandler is part of the ddm site server right? Can we not use a ddm site server?

tests/test_asset_exchange.py
```diff
@@ -0,0 +1,68 @@
...
+    server.server_close()
+    thread.join()
+
+
+def test_asset_download(temp_path, asset_id, image_server, caplog):
+    caplog.set_level(logging.DEBUG)
+    registry_client = MagicMock()
+    empty_ddm = ReplicaUpdate[RegisteredObject](
+            0, 0, datetime(2100, 1, 1, 0, 0, 0), set(), set())
+    registry_client.get_updates_since = lambda: empty_ddm
```
svenvanderburg, line 58: This could be a separate fixture

svenvanderburg (approved):
Nice 👍 I just added some nagging comments on details, but overall I like the new feature!

proof_of_concept/definitions/assets.py
LourensVeen: So far, we've distinguished assets by whether the data field was `None`, but now that we have data asset with an image rather than a value, that no longer works. So I added the attribute, and I may well decide to remove the direct execution, remove the `ComputeAsset` and `DataAsset` classes, and only have `Asset`, but with the `kind` attribute so we can check that someone didn't accidentally try to use a data asset to compute something. Not sure entirely how that will play out, but I think having it is an improvement.

proof_of_concept/definitions/assets.py
LourensVeen: I hate you ;-)

proof_of_concept/rest/client.py
LourensVeen: Hm.
```
if r.headers...:
    chunk_size = None
else:
    chunk_size = CHUNK_SIZE
```
I vaguely recall playing with chunk sizes when I was optimising transfer speed in cerulean, and I don't think the exact size mattered very much there, increasing the number of simultaneous requests in flight was what did the trick. So it's a technical detail that I don't expect to ever need to fiddle with again.

tests/test_asset_exchange.py
LourensVeen: We could, but the DDM site server also serves a bunch of other stuff, so I'd have to mock everything else in the site. This seemed like less work.

tests/test_asset_exchange.py
LourensVeen: Oh, I like that.

tests/test_asset_exchange.py
svenvanderburg: Cool

proof_of_concept/rest/client.py
svenvanderburg: Bueno

proof_of_concept/definitions/assets.py
svenvanderburg: Check

proof_of_concept/definitions/assets.py
svenvanderburg: Haha, sorry :)

PR #57, opened 2021-02-10
Task 1326: execute steps using containers

LourensVeen:
This implements execution of workflow steps by instantiating containers and connecting them via the local Docker network. It's a basic implementation only, and doesn't offer any job isolation or cross-domain functionality.

Note that commit 31ed3e2 was reviewed before as part of a previous PR. I ended up removing the code from there, but now it's back. So that can be skipped, it already incorporates the feedback given before.

8550e19 is a bug fix that's not directly related to execution, it turns out that the local file path was sent rather than the image URL, and it escaped the previous test but was caught by the new one in this PR.


proof_of_concept/components/domain_administrator.py
```diff
@@ -0,0 +1,320 @@
...
+                containers[name] = self._dcli.containers.run(
+                        images['<output>'].id, name=docker_name,
+                        detach=True, network_mode='bridge')
+            return containers
+        except Exception:
+            for container in containers.values():
+                container.remove(force=True)
+            raise
+
+    def _create_config_file(
```
svenvanderburg, line 231: Some repetition of Sven-review-Lourens history here: I know that the code documents itself, but adding a comment like `# Adding input addresses to config` would greatly improve readability here IMO.

proof_of_concept/components/orchestration.py
```diff
@@ -173,7 +173,7 @@ def execute_workflow(self, submission: JobSubmission) -> Dict[str, Any]:
                         results[wf_outp_name] = asset.data
                     except KeyError:
                         continue
-            sleep(0.5)
+            sleep(5)
```
svenvanderburg, line 176: Always nice to sleep a little longer :)

proof_of_concept/components/domain_administrator.py
```diff
@@ -0,0 +1,320 @@
...
+                    container.remove(force=True)
+
+        if compute_container is not None:
+            try:
+                compute_container.remove()
+            except Exception as e:
+                logger.warning(
+                        f'Failed to remove container {compute_container.id}:'
+                        f'{e}')
+                compute_container.remove(force=True)
```
svenvanderburg: This could be a for loop with one block instead of 2 separate blocks.

proof_of_concept/components/domain_administrator.py
```diff
@@ -0,0 +1,320 @@
...
+        if images is not None:
+            for name, image in images.items():
+                try:
+                    self._dcli.images.remove(image.id)
+                except docker.errors.ImageNotFound:
+                    # base output images may already have been deleted
+                    pass
+                except Exception as e:
+                    logger.warning(f'Failed to remove image {image.id}: {e}')
+                    self._dcli.images.remove(image.id, force=True)
```
svenvanderburg: This could be a for loop with one block instead of 2 separate blocks.

proof_of_concept/components/step_runner.py
```diff
@@ -150,9 +136,28 @@ def _is_legal(self) -> bool:
 
         return True
 
+    def _try_step(
```
svenvanderburg:
```suggestion
    def _try_execute_step(
```
?

proof_of_concept/components/step_runner.py
```diff
@@ -150,9 +136,28 @@ def _is_legal(self) -> bool:
...
+            self, step: WorkflowStep, id_hashes: Dict[str, str]
+            ) -> bool:
+        """Try to execute a step, if its inputs are ready."""
```
svenvanderburg: Maybe document that this will try to use docker execution if possible, otherwise run in python.

proof_of_concept/components/step_runner.py
```diff
@@ -187,6 +192,29 @@ def _get_step_inputs(
 
         return step_input_data
 
+    def _run_step(
+            self, step: WorkflowStep, inputs: Dict[str, Asset],
+            compute_asset: ComputeAsset, id_hashes: Dict[str, str]) -> None:
+        """Run a workflow step."""
```
svenvanderburg, line 203:
```suggestion
        """Run a workflow step in python. (As apposed to docker)"""
```

tests/test_containers.py
```diff
@@ -0,0 +1,200 @@
...
+def compute_asset_tar(dcli, docker_dir):
+    compute_file = _build_compute_image(dcli, docker_dir)
+    yield compute_file
+    compute_file.unlink()
+
+
+@patch(
+        'proof_of_concept.components.domain_administrator.OUTPUT_ASSET_ID',
+        Identifier('asset:ns:output_base:ns:test_site'))
+def test_container_step(
```
svenvanderburg, line 121: Really cool that this works!

proof_of_concept/components/domain_administrator.py
```diff
@@ -0,0 +1,320 @@
...
+from proof_of_concept.rest.client import SiteRestClient
+
+
+OUTPUT_ASSET_ID = Identifier('asset:ns:output_base:ns:site')
+
+
+logger = logging.getLogger(__name__)
+
+
+class PlainDockerDA(IDomainAdministrator):
```
svenvanderburg, line 29: I would call this StepRunner or DockerStepRunner or something and not DomainAdministrator, because to me that is what this class does. But that is because I don't know what it means to administrate a domain, so that might be a fancy name for the same thing that I just need to learn...

svenvanderburg (approved):
Wow, nice work! Really cool that this works now :) I left the usual comments, mostly trying to push you to add a little bit more documentation. But I reckon that this code is very much in transition, so feel free to ignore my comments, overall it looks really good!.

proof_of_concept/components/domain_administrator.py
LourensVeen: I'll go expand the docstrings into multi-line ones, you're right that I'm getting a little lazy in this area :-).

proof_of_concept/components/orchestration.py
LourensVeen: Ha! Yeah, with containers everything is slower, and it was producing a lot of uninteresting log output which obscured the interesting stuff. I should really do an exponential drop-off here at some point.

proof_of_concept/components/domain_administrator.py
LourensVeen: Hmm, I guess so. I was mirroring the functions that set everything up in reverse order I guess. Maybe I should make that explicit and split off each block as a separate function? Then I can make a `_remove_containers(self, list)`...I think that's a good idea, thanks!

proof_of_concept/components/domain_administrator.py
LourensVeen: Well, I already have a StepRunner class, so I couldn't use that, and there will be other runners that use Docker to start the containers, but do it in other ways. So I wasn't sure what to call this, and ended up going with the term that the scientists came up with. I think "domain" refers to an administrative domain, e.g. it's the kingdom of the sysadmin/network engineer in charge of the site, and "administrator" means that it controls the things (containers, network connections) in that domain. I'm not super-happy with it, but don't really have a better solution either.

tests/test_containers.py
LourensVeen: Yeah, although it's a bit of a hack. Eventually the compute asset should have some metadata that specifies which base image to use for each output, so that you can have outputs as files (on a  WebDAV) or as triples (in a Virtuoso server or something) or records (in a relational database). And then type declarations for inputs and outputs so that you can check that your workflow ties things together correctly, and ... but we'll probably never get there, and this works.

proof_of_concept/components/domain_administrator.py
LourensVeen: And...implemented.

proof_of_concept/components/domain_administrator.py
LourensVeen: Added some documentation to this effect.

PR #64, opened 2021-03-16
1721: policy management API

LourensVeen:
Let's add more APIs! This is the simplest possible API for managing policies, it only allows you to add rules. That's enough for our experiments, so we'll leave listing and deleting them for some other time.


svenvanderburg (approved):
Looks good!

PR #67, opened 2021-04-13
Task 1722: Workflow submission API

LourensVeen:
This adds an internal REST API that clients can use to submit workflows to their local site, which will then try to run them.

### DIFF
--- a/proof_of_concept/rest/ddm_site.py
+++ b/proof_of_concept/rest/ddm_site.py
@@ -23,7 +23,7 @@ from proof_of_concept.definitions.workflows import JobSubmission
 from proof_of_concept.policy.replication import PolicyStore
 from proof_of_concept.rest.replication import ReplicationHandler
 from proof_of_concept.rest.serialization import deserialize, serialize
-from proof_of_concept.rest.validation import Validator, ValidationError
+from proof_of_concept.rest.validation import site_validator, ValidationError
 
 
 logger = logging.getLogger(__name__)
@@ -145,16 +145,14 @@ class AssetImageAccessHandler:
 
 class AssetManagementHandler:
     """A handler for the internal /assets endpoint."""
-    def __init__(self, store: IAssetStore, validator: Validator) -> None:
+    def __init__(self, store: IAssetStore) -> None:
         """Create an AssetManagementHandler handler.
 
         Args:
             store: The asset store to send requests to.
-            validator: The validator to use on received objects.
 
         """
         self._store = store
-        self._validator = validator
 
     def on_post(self, request: Request, response: Response) -> None:
         """Handle adding an asset.
@@ -166,7 +164,7 @@ class AssetManagementHandler:
         """
         try:
             logger.info(f'Asset storage request')
-            self._validator.validate('Asset', request.media)
+            site_validator.validate('Asset', request.media)
             asset = deserialize(Asset, request.media)
             logger.info(f'Storing asset {asset}')
             self._store.store(asset)
@@ -227,17 +225,14 @@ class AssetImageManagementHandler:
 
 class PolicyManagementHandler:
     """A handler for the internal /rules endpoint."""
-    def __init__(
-            self, policy_store: PolicyStore, validator: Validator) -> None:
+    def __init__(self, policy_store: PolicyStore) -> None:
         """Create a PolicyManagementHandler handler.
 
         Args:
             policy_store: A policy store to store rules in.
-            validator: A validator to validate rule JSON with.
 
         """
         self._policy_store = policy_store
-        self._validator = validator
 
     def on_post(self, request: Request, response: Response) -> None:
         """Handle request to add a rule.
@@ -248,7 +243,7 @@ class PolicyManagementHandler:
 
         """
         try:
-            self._validator.validate('Rule', request.media)
+            site_validator.validate('Rule', request.media)
             rule = deserialize(Rule, request.media)
             self._policy_store.insert(rule)
         except ValidationError:
@@ -259,17 +254,13 @@ class PolicyManagementHandler:
 
 class WorkflowExecutionHandler:
     """A handler for the external /jobs endpoint."""
-    def __init__(
-            self, runner: IStepRunner, validator: Validator
-            ) -> None:
+    def __init__(self, runner: IStepRunner) -> None:
         """Create a WorkflowExecutionHandler handler.
 
         Args:
             runner: The runner to send requests to.
-            validator: A Validator to validate requests with.
         """
         self._runner = runner
-        self._validator = validator
 
     def on_post(self, request: Request, response: Response) -> None:
         """Handle request to execute part of a workflow.
@@ -281,7 +272,7 @@ class WorkflowExecutionHandler:
         """
         try:
             logger.info(f'Received execution request: {request.media}')
-            self._validator.validate('JobSubmission', request.media)
+            site_validator.validate('JobSubmission', request.media)
             submission = deserialize(JobSubmission, request.media)
             self._runner.execute_job(submission)
         except ValidationError:
@@ -316,8 +307,6 @@ class SiteRestApi:
         with open(site_api_file, 'r') as f:
             site_api_def = yaml.safe_load(f.read())
 
-        validator = Validator(site_api_def)
-
         rule_replication = ReplicationHandler[Rule](policy_store)
         self.app.add_route('/external/rules/updates', rule_replication)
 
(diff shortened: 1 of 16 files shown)


proof_of_concept/rest/ddm_site.py
```diff
@@ -290,6 +285,97 @@ def on_post(self, request: Request, response: Response) -> None:
...
+            job = deserialize(Job, request.media)
+            job_id = self._orchestrator.start_job(requester, job)
+            logger.info(f'Received new job {job_id} from {requester}')
+            response.status = HTTP_303
+            response.location = (
+                    f'{request.forwarded_prefix}/internal/jobs/{job_id}')
+        except ValidationError:
+            logger.warning(f'Invalid execution request: {request.media}')
+            response.status = HTTP_400
+            response.body = 'Invalid request'
```
svenvanderburg, line 332:
```suggestion
        try:
            site_validator.validate('Job', request.media)
        except ValidationError:
            logger.warning(f'Invalid execution request: {request.media}')
            response.status = HTTP_400
            response.body = 'Invalid request'
        else:
            job = deserialize(Job, request.media)
            job_id = self._orchestrator.start_job(requester, job)
            logger.info(f'Received new job {job_id} from {requester}')
            response.status = HTTP_303
            response.location = (
                    f'{request.forwarded_prefix}/internal/jobs/{job_id}')
```

svenvanderburg: Maybe use a smaller try block here?

proof_of_concept/rest/validation.py
```diff
@@ -39,3 +41,15 @@ def validate(self, class_: str, user_input: JSON) -> None:
...
+# This is a sort-of singleton, which in this case is actually not a
+# problem. It's not the best possible solution however, so this is
+# temporary. Later, we'll TODO refactor the YAML files and have a
+# plain function as the API of this module.
+
+_site_api_file = Path(__file__).parent / 'site_api.yaml'
+with open(_site_api_file, 'r') as f:
+    _site_api_def = yaml.safe_load(f.read())
+
+site_validator = Validator(_site_api_def)
```
svenvanderburg, line 55: OMG 👀  🤯 A singleton

proof_of_concept/rest/internal_client.py
```diff
@@ -52,3 +59,74 @@ def add_rule(self, rule: Rule) -> None:
...
+        if 'location' not in r.headers:
+            raise RuntimeError('Invalid server response when submitting job')
+        job_uri = r.headers['location']
+        prefix = f'{self._endpoint}/jobs/'
+        if not job_uri.startswith(prefix):
+            raise RuntimeError(
+                    'Unexpected server response when submitting job')
+        return job_uri
+
+    def is_job_done(self, job_id: str) -> bool:
```
svenvanderburg, line 87: Maybe it's a good idea to use `job_uri` here as well instead of `job_id`, because in the WorkflowOrchestrator the job_id is an integer and here it is actually the full uri with which we can get the job.

proof_of_concept/rest/ddm_site.py
```diff
@@ -290,6 +285,97 @@ def on_post(self, request: Request, response: Response) -> None:
...
+
+    This lets internal users submit jobs, which are identified by a
+    unique URI.
+
+    """
+    def __init__(self, orchestrator: WorkflowOrchestrator) -> None:
+        """Create a WorkflowSubmissionHandler handler.
+
+        Args:
+            orchestrator: The orchestrator to send requests to.
```
svenvanderburg: This is somewhat confusing. Because we don't send an HTTP request to the orchestrator. (although we request it to start a job.

svenvanderburg: Same for docstring of the StatusHandler

svenvanderburg (approved):
Nice addition! I notice that the base you have laid out with swagger + falcon + serialization + validation makes an addition like this quite clean, it all still fits well together!

I have added a few small suggestions for improvement.

Also, maybe it's a good idea to add the orchestrator and executor to [terminology](https://github.com/SecConNet/proof_of_concept/blob/develop/docs/terminology.rst)? Just to make clear the sequence of submitting a job, i.e. internal site API -> orchestrator -> executor.

proof_of_concept/rest/ddm_site.py
LourensVeen: I'd prefer having everything at the end, for two reasons: 1) the code isn't interrupted by error handling all the time (a disadvantage of error return codes), and 2) you have an overview of all the possible results at the end of the function.

I think generally the idea is to handle exceptions in the most logical place, which is here just before ending the function and returning the response to the client. On the other hand, if you have two functions which can return the same type of exception object, and it matters which of the functions raised it, then you'll have to split the try block.

One more note: if doing this, I would add a return to the exception handler rather than using an `else`, because if you have multiple potential issues to handle then your column gets a bit narrower after every one. So an early return is better in that case in my opinion.

proof_of_concept/rest/validation.py
LourensVeen: I know. You've corrupted me. But I'd like to think of it as being an expert who knows when to apply the rules and when to break them.

proof_of_concept/rest/internal_client.py
LourensVeen: I was thinking to hide the fact that it's a URI from the user. We could also extract that number here and pass that, which is actually what I had until I realised that it was easier to just pass the URI itself. But I'd like for this to be an opaque handle. Maybe I should call it a job handle (since we also have the Identifier class, which this is definitely not) but anything called handle always sounds so Windowsy... Hmm.

proof_of_concept/rest/ddm_site.py
LourensVeen: Oh, I was thinking request to execute a workflow in general terms, not an HTTP request per se, but it's a fair point. This better?

LourensVeen:
I've added the Orchestrator to the terminology document. I've left out the Executor for now, I think it may be better to call it the Coordinator actually, since it doesn't execute anything itself (that's the StepRunner). I'll leave that for a future documentation improvement.

proof_of_concept/rest/ddm_site.py
svenvanderburg: Interesting, I don't think I have ever encountered someone who prefers this. But I totally agree with your points! Thanks for sharing :)

In my opinion a downside of doing it all at the end is that it is not clear which exceptions can be raised by which parts of the try block. But in this case it is very likely that a method called `validate` might raise this `ValidationError`.

So no need for change here!

proof_of_concept/rest/internal_client.py
svenvanderburg: Hmmm, ok. I think for now this is fine then, I was able to figure out that it was a uri so I guess any other contributor could (considering I am probably the laziest contributor you will ever encounter)

proof_of_concept/rest/validation.py
svenvanderburg: Haha, I think you are actually such an expert!

proof_of_concept/rest/ddm_site.py
svenvanderburg: Better :)

proof_of_concept/rest/ddm_site.py
LourensVeen: The thing is that when an error occurs, there are not that many things you can do. Basically, it boils down to 1) try again (for intermittent faults), 2) try something else (if an alternative exists at this level in the call stack, fairly rare), 3) give up and let someone else deal with it (either calling code or, as last resort, the user).

With exceptions, 3) is the default, they bubble up the stack if you don't do anything. In some cases they need to pop out of the top of the stack in a different way than a printed exception trace, so you need to translate. If you have a reactive system processing requests from the outside world, like a WSGI server, then often the error is logger or returned as the result of the request, and then you move on to the next request. In that case, it's important to make sure that no exception can escape and shut down the system (instant denial-of-service attack there). I actually have a bare `except:` clause in Cerise for that. Gets me in trouble with `pylint`.

In this case, the exception is translated into an HTTP status and passed off to the caller that way. WSGI actually turns any escaping exceptions into 500 Internal Server Error (thus avoiding the DOS), but that's not right for an invalid input, so it needs to be handled separately here and turned into the correct code.

In a well-designed library, you have a nice hierarchy of exception types, maybe with machine-readable additional information as needed rather than just an error message, so that you can catch specific ones if needed or catch a broader category where it's all a wash anyway. Thinking about which of the above categories is likely the way a user of the library would handle the exception can help with designing a good set of exception classes. Or you can just be lazy and raise a `RuntimeError` everywhere, like I too often do :-).

I'd like a better solution for 1) actually. The `retrying` package in Python is okay, but it would be nicer to have a `retry(n times)` command you can put into an `except` clause, or `retry(for n seconds)`. Same effect but better syntax, and more granular because it's at the level of a try block, not a function.

Options 1) and 2) are relatively rare, so even if there are many different things that can go wrong, in many different places, often many of these errors need to be handled in the same way simply because there aren't many options. In that case catching them all at once is nice and concise.

With error returns, you end up explicitly handling possible errors for every single call, often just passing them on up the stack explicitly. That does have the advantage of it being very clear what's going on, and some people like it because of that (like Linus Torvalds, one of his reasons for not switching the kernel to C++). The designers of Go also use return types, but their idea is more that most errors aren't actually all that exceptional, and should just be considered possible results of calling a function, and handled in place. I do like that actually, and their way of returning a `result, error` tuple is cleaner than C's `errno`. I suspect it also lets you do things with those mystical monads. Anyway, they do have an exception system in Go, but it's called `panic` and it's intended for truly exceptional things like not being able to allocate memory (quite hard to do on modern systems). I'm not completely against this, but it gets incredibly tedious sometimes...

LourensVeen:
The tests are timing out because apparently they're creating more network connections than the GitHub Actions infrastructure will allow. It's all passing locally though, so I'm going to merge this and fix the CI separately.

PR #69, opened 2021-04-30
Rename JobSubmission to ExecutionRequest

LourensVeen:
This is a simple rename, although it touches quite a bit of code. I think the CI is still broken, but it passes locally; I'll have a look at that next.


svenvanderburg (approved):
Looks good! Definitely makes things clearer!

PR #70, opened 2021-04-30
Wait for REST servers to come up after starting them

LourensVeen:
There's a race condition in some of our tests, where we start up REST servers but don't wait for them to be up. If there's no contention, then start-up is fast enough that the subsequent connect works, but on the CI there are sometimes periods of heavy contention when everything runs slowly and the connections time out. This adds an attempt to connect to the server right after starting it, with an extra-long timeout of 10 minutes. Hopefully, that should allow the tests to complete even when the CI is slow.

This is hard to test, the tests are passing right now also without this improvement. But it should help, so let's just implement it and hope it fixes things for the future.

### DIFF
--- a/tests/test_asset_exchange.py
+++ b/tests/test_asset_exchange.py
@@ -6,6 +6,7 @@ from wsgiref.simple_server import WSGIRequestHandler
 
 from falcon import App
 import pytest
+import requests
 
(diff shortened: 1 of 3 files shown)


svenvanderburg (approved):
Good that you found this! And solution looks sane!

PR #71, opened 2021-05-02
Task 1258: Split off REST bits from RegistryClient

LourensVeen:
RegistryClient did two things: 1) manage a replica of the registry and allow clients to register callbacks when there were changes, and 2) talk to a registry REST API via requests and the replication system's REST helpers. This PR splits off the second part of its functionality into a new RegistryRestClient class. Note that we already had one of those, which was just a typedef for a REST client for the registry's replication REST API. That's been incorporated into the new class.

Overall, this increases the amount of code and it adds some abstraction boilerplate, but it more cleanly separates the concerns. Uncle Bob would be proud.

Note: please don't merge, others need to go first and this needs to be merged into a different branch.

### DIFF
--- a/proof_of_concept/components/ddm_site.py
+++ b/proof_of_concept/components/ddm_site.py
@@ -32,8 +32,6 @@ class Site:
             rules: List[Rule], registry_client: RegistryClient) -> None:
         """Create a Site.
 
-        Also registers its runner and store in the global registry.
-
         Args:
             name: Name of the site
             owner: Party which owns this site.
--- a/proof_of_concept/components/registry_client.py
+++ b/proof_of_concept/components/registry_client.py
@@ -1,50 +1,41 @@
 """Functionality for connecting to the central registry."""
 from pathlib import Path
-import requests
 from typing import Any, Callable, List, Optional, Set
 
 from cryptography.hazmat.primitives.asymmetric.rsa import RSAPublicKey
-import ruamel.yaml as yaml
 
 from proof_of_concept.definitions.identifier import Identifier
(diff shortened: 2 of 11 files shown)


svenvanderburg (approved):
Nice, it indeed makes sense to factor out the REST client. The documentation that you added also makes it clear how to use this. 

It feels a bit funky that in order to interact with the global registry through REST you need to setup a `RegistryRestClient` and pass that to a `RegistryClient`. I am wondering if we could have really 2 distinct clients: 
(1) `RegistryReplicationClient`: that is the current `RegistryClient` without the forwarding to the `RegistryRestClient`, i.e. it's job is to manage the replication, callbacks, etc.
(2) and a `RegistryRegistrationClient`, which is the current rest client. It's job is to provide an interface for registering and deregistering things. What do you think?

LourensVeen:
Yes, that would effectively split up `IRegistry` into `IReplicationService[RegisteredObject]` and `IRegistration` or something. I think it would make sense, since there may be closed DDMs in which you cannot just register your site through the API, it has to be done with contracts and handshakes and papers with information exchanged with a managing organisation. In that case you'd still need to do the replication, but the registering bit would not apply.

I'll go give this a shot, and I'm thinking to put into this PR rather than making a new one. Does that make sense?

LourensVeen:
Okay, here we go. I like it. Note that RegistryRestClient will become an instantiation of ReplicationRestClient like _ReplicationRestClient is now when we refactor the Validator. So this is still a bit ugly, but will be fixed with that task.

svenvanderburg (approved):
I like it! Ready to merge :)

Issue #65, opened 2021-03-22
Clean up OpenAPI definitions


LourensVeen:
We currently have a `site_api.yaml` which documents the external API of a site. We now have some of the internal API as well, but there's no OpenAPI description of it. We should split the current site_api.yaml into two files, one for the internal and one for the external API, with a third file to contain all the type definitions which is referenced from the other two. The Validator should then be able to just use the third one to do schema checking, with the other files essentially just being documentation.

LourensVeen:
Closed by #72

PR #72, opened 2021-05-03
Implement JSON validation as a function and refactor schemas

LourensVeen:
Implements JSON validation as a function (JVaaF). Thanks to @svenvanderburg for [inventing this](https://github.com/SecConNet/proof_of_concept/pull/36#discussion_r519927186).


LourensVeen:
Oh, this is incomplete with respect to #65, because it doesn't add the endpoint definitions for the internal site APIs. I guess I should go type these up.

proof_of_concept/components/policy_client.py
```diff
@@ -9,7 +9,7 @@
 from proof_of_concept.policy.replication import RuleValidator
 from proof_of_concept.replication import Replica
 from proof_of_concept.rest.replication import PolicyRestClient
-from proof_of_concept.rest.validation import site_validator
+from proof_of_concept.rest.validation import validate_json
```
svenvanderburg: `validate_json` doesn't seem to be used?

svenvanderburg (approved):
Nice, great pattern this VAAF ;) 

As for the swagger yml file changes, I think it is just refactoring `site_api.yaml` into separate internal and external endpoints and the schemas. I agree that that is a nice separation! Just for my interest: Does these separate api specs work nicely with the swagger ui?

LourensVeen:
The internal endpoint wasn't written up yet in an OpenAPI definition, so that is new. Other than that it's just reorganised.

Since the internal and the external endpoints share some of the types, and since it's easier to implement on the Python side, I've put all the type definitions in a single file, and then there are two separate files for the two APIs. But these are really separate APIs, and that separation is very important from a security standpoint. So I want to have them as separate files, and if there's to be a UI, then it should be a separate one for each. I don't know if the Swagger editor can resolve this automatically. I'll try that eventually, as I still need to validate the endpoint descriptions as well.

It's a bit funky anyway in Python, validation is done using the standard Python support library, but that doesn't actually implement the specification correctly. OpenAPI uses JSON Schema, but modifies some things, and the validator doesn't have those modifications. And even references within a single file don't work out of the box, I have to make a resolver that looks them up to make that work. Bit funny, but it works.

proof_of_concept/components/policy_client.py
LourensVeen: Good catch!

PR #74, opened 2021-05-06
Use tox for tests

LourensVeen:
Convert the test system from setuptools calling pytest which then uses plug-ins to call mypy and the linters to using tox to call pytest, mypy and the linters. Also moves documentation building from setuptools to tox, although I'm not sure that that's an improvement per se.

Configuration is now in one file per tool, rather than having everything in setup.py. I think I'd prefer a single file actually, but of course the tools don't agree on which config files to use, so that doesn't work. Maybe in the future `pyproject.toml` could take that role, but it's fairly new and support is still spotty, so that's for the future.


svenvanderburg (approved):
Nice! From the code-complexity perspective this doesn't change a lot, but I think by using tox for this instead of setuptools + github actions you take away a lot of pain!

LourensVeen:
This one got merged, but GitHub got a bit confused about it because I still had the reference branch set wrong. All is well though.

Issue #23, opened 2020-10-28
Improve error handling


LourensVeen:
In many places the current implemention is sloppy. That's okay, since it's only a research prototype, but having better error handling does help even there to find problems quickly, so it would be good at some point to make a pass through the code, think about everything that could go wrong everywhere, and put in some better error handling, more specific exception classes, etc.

PR #76, opened 2021-05-24
Add ComputeMetadata and use it for specifying output assets

### LINKED ISSUE #75: Specify ComputeAsset output assets in metadata
Compute assets that are implemented as containers save their outputs to output containers. Currently, the image to use for those output containers is hardcoded. It should be in the compute asset's metadata.

LourensVeen:
Closes #75.

### DIFF
--- a/proof_of_concept/components/asset_store.py
+++ b/proof_of_concept/components/asset_store.py
@@ -6,7 +6,8 @@ from shutil import copyfile, move, rmtree
 from tempfile import mkdtemp
 from typing import Dict, Optional
 
-from proof_of_concept.definitions.assets import Asset
+from proof_of_concept.definitions.assets import (
+        Asset, ComputeAsset, DataAsset)
 from proof_of_concept.definitions.identifier import Identifier
 from proof_of_concept.definitions.interfaces import IAssetStore
 from proof_of_concept.policy.evaluation import (
@@ -99,17 +100,25 @@ class AssetStore(IAssetStore):
         """
         logger.info(f'{self}: servicing request from {requester} for data: '
                     f'{asset_id}')
-        try:
-            asset = self._assets[asset_id]
+        if asset_id not in self._assets:
+            msg = (
+                    f'{self}: Asset {asset_id} not found'
+                    f' (requester = {requester}).')
+            logger.info(msg)
+            raise KeyError(msg)
+
+        asset = self._assets[asset_id]
+
+        if isinstance(asset, DataAsset):
             perms = self._permission_calculator.calculate_permissions(
                     asset.metadata.job)
             perm = perms[asset.metadata.item]
-            if not self._policy_evaluator.may_access(perm, requester):
-                raise RuntimeError(f'{self}: Security error, access denied'
-                                   f'for {requester} to {asset_id}')
-            logger.info(f'{self}: Sending asset {asset_id} to {requester}')
-            return asset
-        except KeyError:
-            logger.info(f'{self}: Asset {asset_id} not found'
-                        f'(requester = {requester}).')
-            raise
+
+        if isinstance(asset, ComputeAsset):
+            perm = self._policy_evaluator.permissions_for_asset(asset_id)
+
+        if not self._policy_evaluator.may_access(perm, requester):
+            raise RuntimeError(f'{self}: Security error, access denied'
+                               f'for {requester} to {asset_id}')
+        logger.info(f'{self}: Sending asset {asset_id} to {requester}')
+        return asset
--- a/proof_of_concept/components/domain_administrator.py
+++ b/proof_of_concept/components/domain_administrator.py
@@ -13,14 +13,11 @@ from docker.models.containers import Container
 
 from proof_of_concept.components.asset_store import AssetStore
 from proof_of_concept.definitions.assets import (
-        Asset, ComputeAsset, DataAsset, Metadata)
+        Asset, ComputeAsset, DataAsset, DataMetadata)
 from proof_of_concept.definitions.identifier import Identifier
 from proof_of_concept.definitions.interfaces import IDomainAdministrator
 from proof_of_concept.definitions.workflows import Job, WorkflowStep
 from proof_of_concept.rest.client import SiteRestClient
-
-
-OUTPUT_ASSET_ID = Identifier('asset:ns:output_base:ns:site')
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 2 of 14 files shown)


LourensVeen:
Converted back to draft, I realised that I haven't implemented the access check for the output base assets yet. That requires getting at least the metadata for the compute asset, which is interesting because we may not have access to the compute asset. So we may need a separate `/metadata` endpoint for assets, which will let you download metadata for assets you don't have access to. Or something like that.

Anyway, please don't review yet, this needs to be fixed first.

LourensVeen:
Okay, hurrah, I think it works now. I ended up implementing this slightly differently. The base assets are still listed in the metadata, but that is for now purely informational, and they're not used. Instead, the base assets to use are added to the workflow step, which already has the compute asset anyway so that makes sense. It also means that we don't have to download metadata while planning, which is good.

Eventually, there will be a step before the planning stage where we complete the workflow, that could be a point where the base assets are retrieved from the metadata and put into the workflow. (And we'll resolve data products there, but those don't exist yet.) Or maybe the compute asset metadata will evolve to have types instead, and there's some other way of getting base assets for a particular type. We'll cross that bridge when we get to it.

The checks ended up being implemented in three different places. That's not right, and needs to be refactored. It's on the to-do list, we'll keep it as-is for now in the interest of actually making some progress...

proof_of_concept/components/asset_store.py
```diff
@@ -6,7 +6,8 @@
 from tempfile import mkdtemp
 from typing import Dict, Optional
 
-from proof_of_concept.definitions.assets import Asset
+from proof_of_concept.definitions.assets import (
+        Asset, ComputeAsset, DataAsset, DataMetadata)
```
svenvanderburg: Are all these imports used?

proof_of_concept/components/domain_administrator.py
```diff
@@ -127,6 +127,7 @@ def _download_images(
             workdir: Working directory for this job.
             inputs: Step input Asset objects to download.
             compute_asset: The ComputeAsset image to download.
+            output_bases: Step output base assets to download.
```
svenvanderburg, line 140: It took me some time to understand what was ment with 'step output base asset'. The way I understand it is that it is an image in which the output is stored. Maybe you can make the description a bit more verbose?

proof_of_concept/components/step_runner.py
```diff
@@ -129,6 +129,13 @@ def _is_legal(self) -> bool:
 
                 # check that we can access the step's outputs
                 for outp_name in step.outputs:
+                    base_item = '{}.@{}'.format(step.name, outp_name)
+                    if base_item in perms:
+                        base_perms = perms[base_item]
+                        if not self._policy_evaluator.may_access(
+                                base_perms, self._this_site):
+                            return False
```
svenvanderburg, line 137: This (and other copies of this piece of code) will be refactored later right?

proof_of_concept/components/step_runner.py
```diff
@@ -198,6 +207,34 @@ def _get_step_inputs(
...
+
+        Return:
+            A dictionary keyed by output name with corresponding
+            base assets.
+
+        """
+        step_output_bases = dict()
+        for out_name, asset_id in step.outputs.items():
+            if asset_id is None:
+                # Should not happen, test misconfigured?
```
svenvanderburg, line 224: Do you want to do anything with this question? Open an issue maybe? And add a TODO?

proof_of_concept/definitions/assets.py
```diff
@@ -28,6 +32,31 @@ def __init__(self, job: Job, item: str) -> None:
         self.item = item
 
 
+class ComputeMetadata(Metadata):
+    """Stores metadata for stored compute assets.
+
+    Attributes:
+        output_base: The asset id of the empty data image to use for
```
svenvanderburg: I guess this is not the asset id, but a dict mapping the name to the asset id?

proof_of_concept/definitions/workflows.py
```diff
@@ -19,12 +20,23 @@ def __init__(
...
-        self.outputs = outputs
+        self.outputs = dict()   # type: Dict[str, Optional[Identifier]]
+
+        for name, base_asset in outputs.items():
+            if base_asset is None:
+                self.outputs[name] = None
+            elif not isinstance(base_asset, Identifier):
+                self.outputs[name] = Identifier(base_asset)
+            else:
+                self.outputs[name] = base_asset
```
svenvanderburg, line 38: Can the asset still be not an Identifier? Even after you also do this conversion in the ComputeMetaData constructor?

svenvanderburg (approved):
Nice Lourens! I've left a few comments, but in general it looks good! #progress :)

proof_of_concept/components/step_runner.py
LourensVeen: Yes. Eventually we could even get rid of non-container-based execution, which would get rid of the outer `if`, but on the other hand, it's kind of convenient for testing and experiments that aren't about distributed execution per se. (Policy-related stuff, for instance.) But it's silly that there are three places where this is implemented, that has to be refactored.

proof_of_concept/components/step_runner.py
LourensVeen: I'd say this is part of #23.

proof_of_concept/definitions/workflows.py
LourensVeen: Yes, those code paths are independent. ComputeMetaData is made when a compute asset is created, a WorkflowStep is made when a workflow is created. Those are done by different people at different times, in general. Eventually we may want to make it so that the WorkflowStep is filled in automatically from the metadata, but at the moment the steps are made directly, in particular in the test cases where having to write `Identifier(...)` all the time is annoying.

proof_of_concept/components/asset_store.py
LourensVeen: DataMetadata isn't, good catch!

proof_of_concept/definitions/assets.py
LourensVeen: Yes, that's the "for each output" part. I'm kind of counting on the type annotations (and Sphinx' rendering thereof) to explain that it's a dictionary. I could change the sentence order so it reads "For each output, the asset id of the empty data image to use." if that is clearer?

proof_of_concept/components/domain_administrator.py
LourensVeen: I've added a more extensive explanation of the various assets involved in running a step, and what they do.

PR #79, opened 2021-06-15
Task 1723: container improvements

LourensVeen:
Here are some improvements to the Mahiru Docker containers. The setup is now going to be different because we'll have VMs representing actual VMs or physical machines at the sites, rather than having one Docker container per simulated site with the whole simulation in a single docker-compose network.

So the container set-up needed some improvements. I'm going to have separate nginx containers, which is better design and allows having separate endpoints for the internal and external site APIs, so nginx is removed. The images are now smaller, and Gunicorn is configured using a file rather than on the command line, because there were too many options to set.

Commit df5774d is a bit ugly, it 

- moves `docker/` to `src/docker/`
- removes nginx from the containers
- removes the `docker-compose.yml`
- speeds up building by copying less
- adds startup and shutdown messages
- runs Gunicorn as root

It's not that big altogether, but these should have been separate commits, and I'm not sure why they aren't. But I think it's reviewable (actually, the entire PR can be reviewed in one go I think), so I'm leaving it in.

Note that I ended up moving `src/docker/` back to `docker/`; the idea of splitting that was that I could put the Python source into `src/python/mahiru`, and set the Docker build context to `src/`, which would put the contents of `build/` (which are big) outside the build context, which would save a bunch of upload time. But then I went and enabled buildkit and discovered that that was no longer an issue, so I moved it back because having `src/python/mahiru` is non-standard and should be avoided if possible.


docker/mahiru/init.sh
```diff
@@ -15,10 +14,13 @@ trap stop_container SIGTERM
 
 echo 'Installed handler' 1>&2
 
-service nginx start
+export PATH=${PATH}:${HOME}/.local/bin
+gunicorn -c /etc/gunicorn.conf.py --pid ${pid_file} &
 
-su -c "gunicorn --pid ${pid_file} --access-logfile /var/log/gunicorn/access.log --error-logfile /var/log/gunicorn/error.log --capture-output --bind 127.0.0.1:8000 '${wsgi_entry_point}'" mahiru &
+echo 'Started gunicorn' 1>&2
```
svenvanderburg, line 20: This is indeed much cleaner!

svenvanderburg (approved):
Nice changes!

Issue #61, opened 2021-03-14
Enable running experiments across VMs


LourensVeen:
We'd like to have a more realistic experimental set-up, with different sites on different VMs. Each site will run inside a container inside its VM, with various ancillary containers in the same VM. One of those will be an experiment controller, which simulates a user who puts assets into the site, adds rules, and submits workflows.

We already have Site and Registry Docker images, but we don't have internal APIs yet through which local users can administrate and use a site. So those need to be added, then we need an experiment controller, and then we need some Ansible scripts to orchestrate the whole thing.

One open issue is that we'll need to add assets and rules at different sites before we can start submitting workflows. So is there a single experiment controller which has access to the internal interfaces of all the sites, or is there an experiment controller at each site? If the latter, do they coordinate? How? Or maybe we have them repeat the workflow submission until it doesn't fail due to lack of permissions?

LourensVeen:
#62 #63 #64 and #67 implemented the internal APIs, #76 and #78 fixed things that contribute to this, and #79 added a bunch of improvements to the Mahiru Docker images to improve deployment.

PR #80, opened 2021-06-18
Task 1723: improved deployment

LourensVeen:
This makes some improvements needed to be able to run Mahiru in a more realistic, distributed setting.

When running behind a reverse proxy, the outside API endpoint address doesn't match what we're listening on directly. In order to be able to generate correct redirects and URL references, we need to know where the user thinks we are. 63620c0 adds code to get that from the environment, which can be set appropriately by the reverse proxy.

For debugging purposes it's nice if you can set the log level, so ec36135 adds that and configures logging to output to standard out, which is nice when you're running inside of a container because the output will end up being available via `docker logs <container>`

Finally, compute containers need to know where the asset containers are that they get their inputs from and write their outputs to. That was done via a configuration file bind-mounted into the container, but that's inconvenient when the container is going to be run on a different machine. Since it's not a lot of data, 0c90910 just puts the JSON inside an environment variable instead.


LourensVeen:
One more step to implementing #61.

proof_of_concept/rest/ddm_site.py
```diff
@@ -34,6 +34,47 @@
...
+    proxy then the following headers must be set:
+
+        X-Forwarded-Proto: The protocol used, http or https
+        X-Forwarded-Host: The outside host part
+        X-Forwarded-Path: The outside path
+
+    Note that X-Forwarded-Path is very non-standard. Because of our
+    split API setup with two different reverse proxies, the external
+    and internal paths aren't the same, so we need the external path
+    as well to be able to generate correct client-resolvable URLs.
```
svenvanderburg, line 57: This seems complicated and non-standard, but you documented it very well. I am curious if there is a simpler solution to this, or better: move things around a bit so we don't have to do this. I have absolutely 0 ideas for alternatives, but I just wanted to see if you do. Otherwise this is nicely documented and a good solution!

svenvanderburg (approved):
Nice changes. I think it's quite clever to pass the config as json in the environment. I left one comment, but it's just to double check.

proof_of_concept/rest/ddm_site.py
LourensVeen: One option would be to run separate WSGI servers for the internal and external APIs. They would have to communicate with the core however, probably through REST, and then you have the same set-up as we have now except you've implemented your HTTP reverse proxy from scratch. So that doesn't help.

A real alternative would be a data-centric architecture with e.g. a PostgreSQL database containing policies, assets, jobs and replicas, and two WSGI servers (one for each API) talking to it, and to each other via it. That would work and be a more standard design, and it would add the transactional persistent store that would be needed anyway in a production setting. It's a fair amount of work to implement though. Maybe in the future...

svenvanderburg: Fair enough! It is clear that you considered alternatives!

LourensVeen: Yes. There's also the standard `Forwarded` header as a potential alternative to `X-Forwarded-Proto` and `X-Forwarded-Host`, but it's quite a bit more complex and it doesn't include the path either, so it doesn't solve the whole problem and is more difficult to parse. So I went with this.

PR #81, opened 2021-06-29
Task 1723: VM experiments

### LINKED ISSUE #61: Enable running experiments across VMs

LourensVeen:
This finally adds all the Vagrant, Ansible and Docker configuration needed to run a compute-to-data scenario on a set of virtual machines, thus simulating a true distributed data exchange. This closes #61.


docker/assets/compute-asset-base.Dockerfile
```diff
@@ -5,13 +5,7 @@
...
 # /home/mahiru/app.py, and run as the non-root mahiru user.
 
-FROM ubuntu:latest
-
-RUN \
-    apt-get update && \
-    apt-get -y dist-upgrade && \
-    apt-get -y install python3 python3-pip python3-venv python3-wheel && \
-    apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
+FROM python:3.9-slim
```
svenvanderburg, line 8: 👍

docs/distributed_scenarios.rst
```diff
@@ -0,0 +1,332 @@
...
+calls these programs and data sets *assets*. To run our experiment, we'll need
+some of these asset container images. They can be built in the same way:
+
+.. code-block:: bash
+
+  make assets
+
+
+The container images this produces can be found in the `build/` directory, if
+you're interested (but honestly, there's not much to see).
```
svenvanderburg, line 128: Lol

scenarios/common/registry/registry_nginx.conf
```diff
@@ -0,0 +1,54 @@
+# See https://docs.gunicorn.org/en/latest/deploy.html
```
svenvanderburg: For a large part these `nginx.conf` files are the same, maybe you can create a base .conf file and use [include](http://nginx.org/en/docs/ngx_core_module.html#include) to include it in specific .conf files?

docs/distributed_scenarios.rst
```diff
@@ -0,0 +1,332 @@
...
+
+Installing software
+-------------------
+
+The distributed set-up uses `Vagrant <https://vagrantup.com>`_ and `Ansible
+<https://ansible.com>`_ to orchestrate the experiment, and we'll also need
+`Docker <https://docker.com>`_ to build some container images. I'm running this
+on Ubuntu Linux, so that's what these instructions are for. If you have a
+different OS, then you'll probably have to do some figuring-out to get things to
+work.
```
svenvanderburg, line 21: Can we make this a bit more non-linux-user friendly? Maybe add some pointers on where to start (i.e. point to the troubleshooting at the end of the file). And add something like 'If you are stuck, please file an issue so we can help you, see CONTRIBUTING.md'

docs/distributed_scenarios.rst
```diff
@@ -0,0 +1,332 @@
+Running distributed scenarios
```
svenvanderburg, line 1: I love this documentation Lourens! It's really clear, and necessary as well. I think it is a good idea to test whether someone (me?) can reproduce this. I've openend #84 for that.

svenvanderburg, line 1: I take it from you adding these instructions that this scenario is intended to be used by others right? At some places it feels a bit like: 'I've done this experiment, I hope it also works for you, good luck!'. I think it's really subtle, but you could try rewriting it a bit so it feels more like we expect loads and loads of users :) i.e. that it reads more like: 'This is how you should run this experiment'

docs/distributed_scenarios.rst
```diff
@@ -0,0 +1,332 @@
...
+call Ansible to configure (or *provision*) the VMs it has created. Docker is
+used both to deploy Mahiru on the machines and by Mahiru itself to execute
+workflow steps.
+
+If you use Vagrant, then the experiment will be set up locally on a set of
+virtual machines. If you have virtual or physical machines available already,
+then you can skip Vagrant and instead tell Ansible about them, then use Ansible
+directly to install the software and run the experiment.
+
+I'm running this locally, with Vagrant and Ansible, and using ``libvirt`` with
```
svenvanderburg, line 36:
```suggestion
We recommend running this locally, with Vagrant and Ansible, and using ``libvirt`` with
```

docs/distributed_scenarios.rst
```diff
@@ -0,0 +1,332 @@
...
+machines by connecting to them, examining their state, and taking action to
+ensure it matches some desired state described in its configuration. Vagrant can
+call Ansible to configure (or *provision*) the VMs it has created. Docker is
+used both to deploy Mahiru on the machines and by Mahiru itself to execute
+workflow steps.
+
+If you use Vagrant, then the experiment will be set up locally on a set of
+virtual machines. If you have virtual or physical machines available already,
+then you can skip Vagrant and instead tell Ansible about them, then use Ansible
+directly to install the software and run the experiment.
```
svenvanderburg, line 34:
```suggestion
We recommend using Vagrant, then the experiment will be set up locally on a set of
virtual machines. For advanced users, if you have virtual or physical machines available already,
then you can skip Vagrant and instead tell Ansible about them, then use Ansible
directly to install the software and run the experiment.
```

docs/distributed_scenarios.rst
svenvanderburg: I added a few suggestions to change the tone a bit, curious what you think of it!

docs/distributed_scenarios.rst
```diff
@@ -0,0 +1,332 @@
...
+
+
+These settings make Ansible connect to the machine that is being configured
+once, then reuse the connection to do multiple things. This is quite a bit
+faster than reconnecting all the time, because SSH has a lot of connection
+set-up overhead.
+
+With my particular combination of Vagrant and Ansible, I was getting some
+deprecation warnings. Everything seems to work fine however, so I disabled them
+here to cut down on the clutter.
```
svenvanderburg, line 88:
```suggestion
With some combinations of Vagrant and Ansible, you can get
deprecation warnings. Everything seems to work fine however, so 
we disable them here to cut down on the clutter.
```

docs/distributed_scenarios.rst
```diff
@@ -0,0 +1,332 @@
...
+.. code-block:: bash
+
+  vagrant destroy -f
+
+
+How it works
+------------
+
+This section is a bit short for now, but hopefully provides a bit of a start to
+figuring out how all this is put together.
```
svenvanderburg, line 210: I think it's already at 90%, but good to give this disclaimer.

svenvanderburg (approved):
Nice work @LourensVeen! 💪 This is a great milestone!

I don't know much about Vagrant, nor Ansible. But it feels like most of the configuration is there to make stuff work, so if it works, the configuration is good. But me not leaving many comments on those parts doesn't mean a lot.

Instead, my english is better than my Vagrant or Ansible, so I left quite some comments on your very decent documentation ;)

scenarios/common/registry/registry_nginx.conf
LourensVeen: You're right. In fact, I think the registry proxy would run fine with the external site configuration :-).

There may be a problem though with the differences being on the inside of the nested structure. I think to really deduplicate it, I would have to have multiple base files for different parts, i.e. one for the top part, one for the top part within the `http` block, and one for the common bits of the second `server` block. I think that would end up being a bit annoying to understand though. Maybe a bit of duplication is not so bad?

docs/distributed_scenarios.rst
LourensVeen: My intended audience for now is computer science PhD students who have a significant skill set, who are willing to invest some time in figuring things out, and whom I'm probably working with directly. The software quality of this thing isn't anywhere near the point where I would want to start offering support, and then there's the issue that I know nothing at all about virtual machines on MacOS or Windows, so I cannot really help anyway.

So I agree in principle, and the MUSCLE3 documentation has lots of that kind of stuff, but here and now "you'll probably have to do some figuring-out" is just a realistic description of the situation the reader is in...

docs/distributed_scenarios.rst
LourensVeen: Yeah. A future me will probably be doing that, and by then I'll have forgotten how this works, so that will be a reasonable simulation, but an actual other person would be even better. Thanks!

LourensVeen: "I've done this experiment, I hope it also works for you, good luck!" is a pretty good description of the state this software is in, in my perception. I'll come back to this documentation once I actually have some decent unit test coverage and error handling and rephrase things here and there, but for now that would violate truth in advertising regulations ;-).

docs/distributed_scenarios.rst
LourensVeen: I don't really want to recommend Vagrant :-). It also depends on what you want to do, if you want to measure network throughput in the system, then I don't recommend doing that on one machine with Vagrant-launched VMs, because that would not be very realistic. So it depends quite a bit on what the user wants to do.

docs/distributed_scenarios.rst
LourensVeen: Same here. I'm actually not explaining how to do this with bare Ansible, without Vagrant. I may try this on the UvA physical machines at some point, so I'll figure that out then and add it.

scenarios/common/registry/registry_nginx.conf
svenvanderburg: Check, that sounds more complicated indeed

docs/distributed_scenarios.rst
svenvanderburg: OK, I think I understand your motivation for writing it this way. I still disagree ;) I'm exaggerating, but the way you see it is that a user-friendly (in the sense of inviting people to use the software instead of scaring away) immediately promises a finished, reliable product. You can also have user-friendly documentation and have a disclaimer somewhere that it is still a prototype. The code quality and documentation of mahiru is much better than some software out there that people use everyday in production.

svenvanderburg: This is my reply also to the other comments, and I think this is a huge improvement over the old code, so feel free to merge! 🚀

scenarios/common/registry/registry_nginx.conf
LourensVeen: Actually, there's a better solution. I'm already fixing up files here and there with Ansible, so I've done that here as well now. There's a single nginx.conf.tmpl, and Ansible copies that three times and then patches it to create the correct file. That has the disadvantage that the whole Ansible thing is not super-transparent, but it is a bit more compact. I'm going to merge it like this, we can always change it again later if it turns out not to work.

PR #90, opened 2021-07-07
Compare replicable objects by value

### LINKED ISSUE #85: Make replicable objects compare by value
This makes it so that when you delete all rules and load in a new set, anything that's the same again won't be replicated as a delete-and-recreate but just not at all. And it's semantically correct, as rules are values.

While we're at it, maybe we should have CanonicalStore treat objects by value, so that if you insert something that's identical to an object already in there, it's a no-op? Currently if you insert a rule twice, you get two copies of it. That doesn't change the semantics, but it's messy and less efficient.

I think Python has an `__equals__` or something, we should look up the semantics and see how that works with `==` and `is` and then implement this.

LourensVeen:
Currently, rules and also registered objects are objects, meaning that you can have for example two rules with exactly the same attributes but a different identity. That doesn't make much sense for rules, nor for sites and parties. (Two different parties with the same name? Two different sites with the same endpoint?)

So this PR modifies the Rule and RegisteredObject classes to compare objects of those classes by their attributes, rather than by their address (which is the default).

This also adds another small change, in that it deduplicates replica updates. Those have a set of objects to remove, and a set of objects to add to the replica to update it from one version to another. Currently, if an object is removed and then re-added to the database in between those two versions, then it ends up in both sets, so it gets transmitted twice and then removed and re-added in the replica as well. That's not wrong, but if we deduplicate the update before sending it then we can save ourselves some trouble. And of course that only works if we compare by value, because the two different versions are different objects in Python.

This does not implement the deduplicate-on-insert mentioned in #85. I decided against it again, because adding an object that already exists is an error, and rather than silently ignoring it I'd rather have the duplicate object in the database where we can see it. We may want to have a separate `overwrite()` operation at some point, but that needs more thought and a sensible use case first.

This closes #85.

### DIFF
--- a/tests/test_rule_replication.py
+++ b/tests/test_rule_replication.py
@@ -1,0 +1,54 @@
+from copy import copy
+
+from proof_of_concept.policy.replication import PolicyStore
+from proof_of_concept.policy.rules import (
+        InAssetCollection, InPartyCollection, MayAccess, ResultOfDataIn,
+        ResultOfComputeIn)
+from proof_of_concept.replication import ReplicableArchive
+from proof_of_concept.definitions.policy import Rule
+
+
+def test_rules_are_values():
+    # If a rule in a policy store is removed and then reinserted, the
+    # reinserted version of the rule should be considered the same
+    # rule, and if a replica update is requested from a version in
+    # which the rule was present to a version in which it is again
+    # present, then the rule should not show up in the update at all.
+
+    # This test runs a small scenario in which a rule is inserted and
+    # deleted and reinserted amidst some other changes, and checks that
+    # it is absent from an update straddling the deletion and
+    # reinsertion.
+    archive = ReplicableArchive[Rule]()
+    policy_store = PolicyStore(archive, 1.0)
+
+    rule1 = InAssetCollection(
+            'asset:party1_ns:data1:ns:s',
+            'asset_collection:party1_ns:collection1')
+    rule2 = InPartyCollection(
+            'party:party2_ns:party2', 'party_collection:party2_ns:collection2')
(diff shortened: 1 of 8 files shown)


LourensVeen:
I added a few stray changes that had been littering my git workspace for way too long, sorry :-).

tests/test_rule_replication.py
```diff
@@ -0,0 +1,44 @@
...
+
+from proof_of_concept.policy.replication import PolicyStore
+from proof_of_concept.policy.rules import (
+        InAssetCollection, InPartyCollection, MayAccess, ResultOfDataIn,
+        ResultOfComputeIn)
+from proof_of_concept.replication import ReplicableArchive
+from proof_of_concept.definitions.policy import Rule
+
+
+def test_rules_are_values():
```
svenvanderburg, line 11: Maybe add a short description of what this test exactly tests, and maybe the playbook. I know I can read the code to figure it out, but I think in this case it helps to quickly grasp it.

svenvanderburg: I guess this comment also counts for the other `test_..._are_values`

svenvanderburg (approved):
Nice! I always love these solutions changing `__eq__`, I think it's really elegant!

You could think of creating a CompareByValue mixin (or parent) class that implements this instead of having the code copied in 3 different cases (I didn't spot differences, but maybe I missed something).

tests/test_rule_replication.py
LourensVeen: I've added an explanation, I agree that especially this one is a bit convoluted as I threw in some background changes as well, to make sure I covered all the rule types.

LourensVeen:
Good point about factoring out `ComparesByValue`, I guess I'm still not completely used to dynamic languages... :-)

PR #91, opened 2021-07-07
Rename package

LourensVeen:
This renames the `proof_of_concept` package to `mahiru` (finally!).

It also adds a `make docs` target to the main Makefile to make building the documentation a bit easier (although there isn't actually any, but I wanted to check that sphinx ran without errors and it turns out we didn't have a good way of building the docs).

Finally, it adds a couple paragraphs of text to the README explaining what Mahiru is and does.

I don't think this is really all that reviewable; it passes all the tests and I've checked the distributed scenarios as well offline. I'll go and rename the actual repository once this is merged.


README.rst
```diff
@@ -1,22 +1,50 @@
...
+and the system will determine automatically whether the workflow can be
+executed, and if so how.
+
+Mahiru's execution model is very flexible, and can support data downloads,
+compute-to-data, software-as-a-service, infrastructure-as-a-service, trusted
+third parties, and distributed machine learning, and it will automatically
+assemble any combination of these required to fulfil a request.
+
+Mahiru is a federated design, which means that participating parties can run a
+Mahiru site on their own hardware, on their own premises and fully under their
```
svenvanderburg, line 33: Wow, really great description Lourens :) I think this is the exact right tone to 'advertise' mahiru.

svenvanderburg (approved):
I scrolled through all the changes in 0.3 seconds. My brain processed lots of renaming from proof_of_concept into mahiru, so that looks good.

Also really like the description you added to the README! You're really taking big steps into creating a good story around this prototype👍

PR #92, opened 2021-10-19
Issue 88: 1. Refactoring and fixes

LourensVeen:
Here's an initial set of changes to implement container connections. They don't actually implement container connections, they just do some refactoring to make that easier, and fix a few small issues.

For d0a0633 it's probably easier to just read the new version of the file, the diff is messy...

### DIFF
--- a/mahiru/components/domain_administrator.py
+++ b/mahiru/components/domain_administrator.py
@@ -30,21 +30,19 @@ class StepResult(IStepResult):
     wrapped up in this class so that we can add a utility function.
 
     Attributes:
-        assets: Dictionary mapping workflow items to Asset objects, one
-                for each output of the step. Each Asset object points
-                to an image file.
+        files: Dictionary mapping workflow output items to Paths of
+                their image files.
     """
-    def __init__(self, assets: Dict[str, DataAsset], workdir: str) -> None:
+    def __init__(self, files: Dict[str, Path], workdir: str) -> None:
         """Create a StepResult.
 
         Args:
-            assets: The output assets created for each step output,
-                    indexed by item and pointing to an image file
-                    on disk.
+            files: Dictionary mapping workflow output items to Paths
+                    of their image files.
             workdir: The working directory in which the image files
                     are stored.
         """
-        self.assets = assets
+        self.files = files
         self._workdir = workdir
 
     def cleanup(self) -> None:
@@ -75,15 +73,20 @@ class PlainDockerDA(IDomainAdministrator):
         # Note: this is a unique ID per executed step, it is unrelated
         # to Job objects, which represent an entire submitted workflow.
         # It is used to avoid name collisions among Docker containers
-        # and images in the docker repository.
+        # and images inside the docker repository.
         self._next_job_id = 1
         self._job_id_lock = Lock()
+
+        # Images currently loaded into Docker, indexed by asset id,
+        # and a reference count.
+        self._loaded_images_lock = Lock()
+        self._loaded_images = dict()            # type: Dict[str, Identifier]
+        self._loaded_images_ref_count = dict()  # type: Dict[str, int]
 
     def execute_step(
             self, step: WorkflowStep, inputs: Dict[str, Asset],
             compute_asset: ComputeAsset, output_bases: Dict[str, Asset],
-            id_hashes: Dict[str, str],
-            step_subjob: Job) -> StepResult:
+            id_hashes: Dict[str, str], step_subjob: Job) -> StepResult:
         """Execute the given workflow step.
 
         Executes the step by instantiating data and compute containers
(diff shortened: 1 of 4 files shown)


mahiru/components/domain_administrator.py
```diff
@@ -31,31 +61,32 @@ class PlainDockerDA(IDomainAdministrator):
...
-        # and images in the docker repository.
+        # and images inside the docker repository.
         self._next_job_id = 1
         self._job_id_lock = Lock()
 
+        # Images currently loaded into Docker, indexed by asset id,
+        # and a reference count.
+        self._loaded_images_lock = Lock()
+        self._loaded_images = dict()        # type: Dict[str, Identifier]
+        self._loaded_images_refs = dict()   # type: Dict[str, int]
```
svenvanderburg:
```suggestion
        self._loaded_images_ref_count = dict()   # type: Dict[str, int]
```

mahiru/components/domain_administrator.py
```diff
@@ -79,142 +110,162 @@ def execute_step(
...
 
-        This downloads the container images for step inputs, compute
-        asset and step outputs from their source asset stores, and
-        saves them as tarballs in the working directory.
+            if assets is not None:
+                for asset in assets.values():
+                    self._free_image(asset.id)
+
+    def _ensure_image_available(self, asset: Asset, workdir: Path) -> Image:
+        """Ensures the asset's image is available in Docker.
```
svenvanderburg, line 174: I think next to ensuring that it's there, it also retrieves it? Maybe update the naming/documentation of this function (and it's plural form)?

svenvanderburg: Or is this because in a later PR they are not retrieved anymore?

mahiru/components/domain_administrator.py
```diff
@@ -79,142 +110,162 @@ def execute_step(
...
-                with path.open('rb') as f:
-                    images[name] = self._dcli.images.load(f.read())[0]
-            return images
+            for name, asset in assets.items():
+                images[name] = self._ensure_image_available(asset, workdir)
+
         except Exception:
-            for image in images.values():
-                self._dcli.images.remove(image.id)
+            for name in images:
```
svenvanderburg, line 260: Can you add a comment here to explain why it makes sense (which it probably does) to free all images in case of an exception. (i.e. in case any of the images is not there, we free all images).

mahiru/components/domain_administrator.py
```diff
@@ -226,13 +277,38 @@ def _start_data_containers(
...
 
+        except Exception:
+            for container in containers.values():
+                container.remove(force=True)
+            raise
+
+    def _start_output_containers(
+            self, job_id: int, images: Dict[str, Image],
+            outputs: Iterable[str]) -> Dict[str, Container]:
+        """Start input and output data containers.
```
svenvanderburg:
```suggestion
        """Start output data containers.
```

mahiru/components/domain_administrator.py
```diff
@@ -226,13 +277,38 @@ def _start_data_containers(
...
+
+        Creates and starts containers for the inputs and outputs of
```
svenvanderburg: Should also be udpated.

mahiru/components/domain_administrator.py
```diff
@@ -444,30 +433,15 @@ def _remove_containers(self, containers: Iterable[Container]) -> None:
...
+                container.stop()
+            except Exception:
+                # ignore, we may be cleaning up an error situation
+                pass
+
             try:
                 container.remove()
             except Exception as e:
                 logger.warning(
                         f'Failed to remove container {container.id}: {e}')
```
svenvanderburg: This warning might be deceiving. Because we use the force in the next step to remove it anyway. Would maybe be better to say: 'Failed to remove container safely, removing with force now'

mahiru/definitions/interfaces.py
```diff
@@ -165,6 +165,27 @@ def execute_request(
...
+    """Contains and manages the outputs of a step.
+
+    Some cleanup is needed after these have been stored, so they're
+    wrapped up in this class so that we can add a utility function.
+
+    Attributes:
+        files: Dictionary mapping workflow items to Path objects, one
+                for each output of the step. Each Path object points
+                to an image file.
+    """
```
svenvanderburg, line 178: This suggests there can be multiple outputs, but the `_save_output` method of `PlainDockerDA` only stores one output. Did I miss something?

svenvanderburg (approved):
Nice changes Lourens! This is a good chunk to review, thanks for feeding it to me in this way 🙏 I left some minor comments, mostly clarifications that would ideally be answered in comments/naming. Of course feel free to do with them what you think is right!

mahiru/components/domain_administrator.py
LourensVeen: Bit long, but you're right that it's clearer enough to be worth it.

mahiru/components/domain_administrator.py
LourensVeen: No, they're being retrieved. I'll expand the docstring a bit, I think the function name is long enough as is :-).

mahiru/components/domain_administrator.py
LourensVeen: If any of the images cannot be made available, then we undo all the work we've done already before we raise the exception so that the Docker daemon is either in the desired state (on success) or in its previous state (on error). This is known as strong exception safety in the C++ world, and it's important to ensure that the system keeps running correctly even if errors occur. I'm not making any such promises on that about this Mahiru implementation, but at least this function gets it right :-). (Weak exception safety, in case you were wondering, leaves the system/object in some well-defined/consistent state, but not necessarily the original one.)

I'll add a brief comment.

mahiru/components/domain_administrator.py
LourensVeen: Good point!

mahiru/definitions/interfaces.py
LourensVeen: `_save_output` is called multiple times in `execute_step`, once for each output, and then all those results are gathered into a single StepResult object. So this is correct. I guess I could make a `_save_outputs` function but if all it does is wrap one loop then it's not really worth it I think.

svenvanderburg:
OK, sorry I missed the loop. Indeed makes sense to not have such a function!


[https://esciencecenter.nl/wp-content/themes/raadhuis/dist/assets/img/e-signature.png]
Sven van der Burg (he, him, his)
eScience Research Engineer/Instructor
My working days are Monday, Tuesday, Wednesday, and Thursday.
Netherlands eScience Center
www.esciencecenter.nl<https://www.esciencecenter.nl/>
________________________________

LourensVeen:
I've implemented all of the suggestions, so this one is getting merged.

PR #93, opened 2021-10-19
Issue 88: 2. Add connections configuration

LourensVeen:
This factors out the Settings class into a separate module, and then adds a section for configuring container connections.

Please don't merge, the base branch is wrong because this is one of my famous stacked PRs.

### DIFF
--- a/mahiru/components/settings.py
+++ b/mahiru/components/settings.py
@@ -1,11 +1,52 @@
 """Site configuration."""
 from pathlib import Path
-from typing import AnyStr, IO, Union
+from typing import AnyStr, IO, List, Optional, Union
 
 import ruamel.yaml as yaml
 import yatiml
 
 from mahiru.definitions.identifier import Identifier
+
+
+class NetworkSettings:
+    """Whether and how to serve remote connections to assets.
+
+    Attributes:
+        enabled: Whether to allow remote connections to assets.
+        external_ip: External IPv4 address we are available at.
+        ports: Port range to use for serving incoming connections,
+            as a length-2 list of the form [min, max].
+    """
+    def __init__(
+            self, enabled: bool = False, external_ip: Optional[str] = None,
+            ports: Optional[List[int]] = None) -> None:
+        """Create a ConnectionSettings object.
+
+        Args:
+            enabled: Whether to allow remote connections to assets.
+            external_ip: External IPv4 address we are available at.
+            ports: Port range to use for serving incoming connections,
+                as a length-2 list of the form [min, max].
+        """
+        if enabled:
+            if not external_ip:
+                raise RuntimeError('External IP not specified')
+            for num in external_ip.split('.'):
+                if int(num) < 0 or int(num) > 255:
+                    raise RuntimeError('Invalid external IP address')
+
+            if not ports:
+                raise RuntimeError('No external port range specified')
+
+            if len(ports) != 2:
+                raise RuntimeError('Expected port range like [min, max]')
+
+            if ports[0] > ports[1]:
+                raise RuntimeError('Minimum must be <= maximum')
+
+        self.enabled = enabled
+        self.external_ip = external_ip
+        self.ports = ports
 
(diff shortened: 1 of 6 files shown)


mahiru/components/settings.py
```diff
@@ -0,0 +1,126 @@
...
+from pathlib import Path
+from typing import AnyStr, IO, List, Optional, Union
+
+import ruamel.yaml as yaml
+import yatiml
+
+from mahiru.definitions.identifier import Identifier
+
+
+class AssetConnections:
```
svenvanderburg:
```suggestion
class AssetConnectionSettings:
```
?? To make it clear this is settings?

mahiru/components/settings.py
```diff
@@ -0,0 +1,126 @@
...
+    Attributes:
+        enabled: Whether to allow remote connections to assets.
+        external_ip: External IPv4 address we are available at.
+        ports: Port range to use for serving incoming connections,
+            as a length-2 list of the form [min, max].
+    """
+    def __init__(
+            self, enabled: bool = False, external_ip: Optional[str] = None,
+            ports: Optional[List[int]] = None) -> None:
+        """Create a ConnectionSettings object.
```
svenvanderburg, line 23: Still refer to ConnectionSettings

mahiru/components/settings.py
svenvanderburg: Or you could refer to it as `settings.AssetConnections`

mahiru/components/settings.py
```diff
@@ -0,0 +1,126 @@
...
+
+            if ports[0] > ports[1]:
+                raise RuntimeError('Minimum must be <= maximum')
+
+        self.enabled = enabled
+        self.external_ip = external_ip
+        self.ports = ports
+
+
+class Settings:
```
svenvanderburg:
```suggestion
class SiteSettings:
```

svenvanderburg (approved):
Love it! This is really clean now 👍 I left a few minor comments.

mahiru/components/settings.py
LourensVeen: Yeah, I'm not very happy with the naming here either. Maybe `NetworkSettings`? That's shorter than `AssetConnectionSettings` but still covers the contents.

mahiru/components/settings.py
LourensVeen: I like adding `Site`, and I think I might prefer `SiteConfiguration` actually. After all, it represents the entire config file, and that's what you usually call them.

LourensVeen:
Okay, they're now SiteConfiguration and NetworkSettings. I'd like to hear what you think.

Also, are you sure you don't want to suggest that the settings should be a singleton/global object?

svenvanderburg:
NetworkSettings​ sounds good!


[https://esciencecenter.nl/wp-content/themes/raadhuis/dist/assets/img/e-signature.png]
Sven van der Burg (he, him, his)
eScience Research Engineer/Instructor
My working days are Monday, Tuesday, Wednesday, and Thursday.
Netherlands eScience Center
www.esciencecenter.nl<https://www.esciencecenter.nl/>
________________________________

svenvanderburg (approved):

svenvanderburg:
> Okay, they're now SiteConfiguration and NetworkSettings. I'd like to hear what you think.
> Also, are you sure you don't want to suggest that the settings should be a singleton/global object?
I thought I could get away with responding to Github emails only, but that clearly doesn't work!

I really like the new names!
And about singletons: In this case it seems like the SiteConfiguration and NetworkSettings are only used in the `Site` object (I only had a quick look though), which makes sense because you need those settings to setup the site correctly. So I would argue against a singleton. If it is (or becomes) more like a global configuration then indeed it would make sense to create a singleton.

LourensVeen:
Ah, merged it but forgot to adjust the reference branch here, so GitHub didn't detect that. Closing.

PR #94, opened 2021-10-19
Issue 88: 3. Add a pilot container

LourensVeen:
This adds a pilot container whose namespace the compute asset container will use. It's not actually used yet, but a future PR will set up the network connections in this pilot container so that it can do it before starting the compute asset.

More stacking here, please don't merge.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -4,3 +4,4 @@ include README.rst
 include mahiru/rest/schemas.yaml
 include mahiru/rest/site_api.yaml
 include mahiru/rest/registry_api.yaml
+include mahiru/data/*
--- a/Makefile
+++ b/Makefile
@@ -5,13 +5,13 @@ docs:
 	PYTHONPATH=${PYTHONPATH}:${PWD} make -C docs html
 
 .PHONY: docker_images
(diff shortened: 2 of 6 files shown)


svenvanderburg (approved):
Nice! Only thing I can remark is: the name 'pilot' was not clear immediately, although when reading the documentation it made sense and it is easy to remember. It could be that these things are always called pilots and I just never heard of it. Alternatives could be something like: `init` or `namespace_provider`.

LourensVeen:
Ah, it's a term from high performance computing, which in turn got inspiration from the pilot light in your gas-powered water heater. In HPC, a pilot job is a small program that runs on a compute note and launches the actual programs you want to run on demand, just like the pilot light lights up the big flame. So here it's a small program that does nothing. I think `init` is too generic and `namespace_provider` may be too specific, I think there's a chance that this container will do some more set up in the future as well. So I'm going to stick with pilot.

svenvanderburg:
OK, that makes sense! I think this is a good case where a bit of jargon actually makes it clear for everyone (that know the jargon of course) what you are talking about. And this one is easy to remember.

LourensVeen:
Merged.

PR #95, opened 2021-10-19
Issue 88: 4. Add net admin helper

LourensVeen:
This adds a clone-and-build script that will create a Docker image in a tarball containing net-admin-helper, configured for WireGuard container networking.

This is of course the perfect use case for git submodules, but I decided against them after discovering that they really only work well if you configure your git with some specific settings, and that the default will leave users unfamiliar with submodules very confused.

One more for the stack

### DIFF
--- a/net-admin-helper/install.sh
+++ b/net-admin-helper/install.sh
@@ -1,0 +1,11 @@
+if [ -d net-admin-helper ] ; then
+    cd net-admin-helper && git checkout develop && git pull && cd ..
+else
+    git clone -b develop https://github.com/SecConNet/net-admin-helper.git
+fi
+
+cp config.h net-admin-helper/
+cd net-admin-helper
+make docker
+docker save net-admin-helper:latest | gzip -1 -c >../../mahiru/data/net-admin-helper.tar.gz
+
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,7 @@ docker_images: base_docker_image registry_docker_image site_docker_image pilot_i
 # faster because the tarball of the first doesn't end up in the build
 # context of the second.
 .PHONY: docker_tars
-docker_tars: docker_images registry_docker_tar site_docker_tar pilot_tar
+docker_tars: docker_images registry_docker_tar site_docker_tar pilot_tar net_admin_helper_tar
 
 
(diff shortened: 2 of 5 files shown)


net-admin-helper/install.sh
```diff
@@ -0,0 +1,11 @@
+if [ -d net-admin-helper ] ; then
+    cd net-admin-helper && git checkout develop && git pull && cd ..
+else
+    git clone -b develop https://github.com/SecConNet/net-admin-helper.git
```
svenvanderburg, line 4: Installing the `develop` branch is maybe not the most sustainable solution (although at this stage the perfect solution I don't know). Maybe it's a good idea to add versioning to `net-admin-helper`, so we can install a specific version or the latest stable version.

svenvanderburg (approved):
Makes sense to me to install `net-admin-helper` in this way. At least it is easy to understand what's going on. An alternative would be to create a github-action for `net-admin-helper` that will deploy a new docker image to github docker registry upon a new release. But this is also totally fine! I left one small comment :)

net-admin-helper/install.sh
LourensVeen: Yes, the plan is to at some point make a release of both, and then the release (main) branch of mahiru will point to a specific release (tag) of `net-admin-helper`. While developing we're on the develop branch here and refer to the develop branch there. If both of these were actively developed by independent teams then that would probably break things regularly, but I think we should be fine here.

LourensVeen:
I considered doing it via Dockerhub, but decided against it because I want Mahiru to be fully cloud-optional. "You can't install Mahiru if Dockerhub is down" is not great from a digital sovereignty perspective. Of course you can say the same thing now for GitHub, so at some point we'll make a VM image with everything baked in to really solve that problem.

svenvanderburg:
> I considered doing it via Dockerhub, but decided against it because I want Mahiru to be fully cloud-optional. "You can't install Mahiru if Dockerhub is down" is not great from a digital sovereignty perspective. Of course you can say the same thing now for GitHub, so at some point we'll make a VM image with everything baked in to really solve that problem.

I am not surprised you already considered it :) Sounds sane!

PR #97, opened 2021-10-21
Issue 88: 6. Domain Administrator serves assets

LourensVeen:
This adds code to the NetworkAdministrator for making and configuring WireGuard endpoints, and then adds code to the DomainAdministrator to use that to serve assets.

Note to self: please fix reference before merging so it auto-closes...

### DIFF
--- a/mahiru/components/domain_administrator.py
+++ b/mahiru/components/domain_administrator.py
@@ -14,6 +14,8 @@ from docker.models.containers import Container
 
 from mahiru.definitions.assets import (
         Asset, ComputeAsset, DataAsset, DataMetadata)
+from mahiru.definitions.connections import (
+        ConnectionInfo, ConnectionRequest)
 from mahiru.definitions.identifier import Identifier
 from mahiru.definitions.interfaces import (
         IDomainAdministrator, INetworkAdministrator, IStepResult)
@@ -78,7 +80,7 @@ class PlainDockerDA(IDomainAdministrator):
         # Note: this is a unique ID per executed step, it is unrelated
         # to Job objects, which represent an entire submitted workflow.
         # It is used to avoid name collisions among Docker containers
-        # and images inside the docker repository.
+        # and images inside the local docker repository.
         self._next_job_id = 1
         self._job_id_lock = Lock()
 
@@ -87,6 +89,12 @@ class PlainDockerDA(IDomainAdministrator):
         self._loaded_images_lock = Lock()
         self._loaded_images = dict()            # type: Dict[str, Identifier]
         self._loaded_images_ref_count = dict()  # type: Dict[str, int]
+
+        # These are indexed by connection id, which is the stringified
+        # job id of the serving job.
+        self._served_lock = Lock()          # Protects the below
+        self._served_assets = dict()        # type: Dict[str, Identifier]
+        self._served_containers = dict()    # type: Dict[str, Container]
 
     def execute_step(
             self, step: WorkflowStep, inputs: Dict[str, Asset],
@@ -121,9 +129,7 @@ class PlainDockerDA(IDomainAdministrator):
             cleanup() on the result after saving the assets to properly
             free the resources.
         """
-        with self._job_id_lock:
-            job_id = self._next_job_id
-            self._next_job_id += 1
+        job_id = self._get_job_id()
 
         logger.info(f'Executing container step {step} as job {job_id}')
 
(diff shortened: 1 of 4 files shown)


mahiru/components/domain_administrator.py
```diff
@@ -180,6 +186,88 @@ def execute_step(
...
+        Args:
+            asset: A local asset to serve.
+            connection_request: A description of the remote end of the
+                VPN connection to set up.
+        """
+        if asset.image_location is None:
+            raise RuntimeError(
+                    f'Cannot serve asset {asset} because it has no image')
+
+        if asset.image_location.startswith('http:'):
```
svenvanderburg, line 206: Does this need to be fool-proof? I can imagine other image locations that are remote, but don't start with 'http' (like ones starting with 'https').

mahiru/components/domain_administrator.py
```diff
@@ -180,6 +186,88 @@ def execute_step(
...
+                del self._served_assets[conn_id]
+            if conn_id in self._served_containers:
+                del self._served_containers[conn_id]
+            raise
+
+    def stop_serving_asset(self, conn_id: str) -> None:
+        """Stop and remove an asset container being served remotely.
+
+        The conn_id argument must have been returned by serve_asset()
+        previously. Implementation note: it's just the job id converted
```
svenvanderburg, line 249: If there is a one-to-one mapping between job and connection, also conceptually, you might as well get rid of the `conn_id` and use `job_id` everywhere. You then start/stop the connection **related to** `job_id`=SOME_JOB_ID.

mahiru/components/network_administrator.py
```diff
@@ -46,7 +71,21 @@ def serve_asset(
         Return:
             A connection info object to send back to the client.
         """
-        raise NotImplementedError()
+        if not isinstance(request, WireGuardConnectionRequest):
```
svenvanderburg, line 74: Maybe update the type hint for `request` as well (that is `ConnectionRequest`)

svenvanderburg: Or will this support more request types in theory?

mahiru/components/network_administrator.py
```diff
@@ -86,3 +132,100 @@ def disconnect_inputs(self, job_id: int, inputs: Dict[str, Asset]) -> None:
...
+    def _ensure_net_admin_helper(self) -> None:
+        """Ensures that the NAH Docker image is loaded into Docker."""
+        try:
+            self._dcli.images.get(_NAH_IMAGE)
+        except docker.errors.ImageNotFound:
+            image_file = (
+                    Path(__file__).parents[1] / 'data' /
+                    'net-admin-helper.tar.gz')
+            with image_file.open('rb') as f:
+                image = self._dcli.images.load(f.read())[0]
```
svenvanderburg, line 208: I have seen similar code like this before. You could have a helper function: `ensure_image_loaded(image_path)` and reuse that throughout your codebase.

mahiru/definitions/connections.py
```diff
@@ -16,3 +16,73 @@ class ConnectionInfo:
...
+    """Describes a request to provide a connection to a data source.
+
+    There are always two sides to a connection, one which initiates
+    connection setup and one which responds. The initiator is the
+    client side of the REST request, and the container on that side
+    contains a compute asset which is the client side of the connection
+    within the VPN tunnel. The responder hosts the REST API for setting
+    up connections, and the container on that side contains a data
+    asset which is the server side of the connection within the VPN
+    tunnel.
```
svenvanderburg, line 52: Is there any documentation missing that is specific to `WireGuardConnectionRequest`?

svenvanderburg (approved):
Looks good! I left some 'could-have' comments.

mahiru/components/domain_administrator.py
LourensVeen: Good point, there will be `https` at least in the future (and `http` needs to go, being insecure). Maybe it would be better to start any local paths with `file://`, test for that, and assume anything else is remote. Or maybe we should even have different Asset classes, one for communication between the sites with a URL and another one for internal use with a local path. That may be better, but doesn't belong in this PR. I've made https://github.com/SecConNet/mahiru/issues/102 for this so I can come back to it later.

mahiru/components/domain_administrator.py
LourensVeen: Okay, this is my fourth attempt at writing an explanation of what's going on. Let's see if it goes better. When our site executes a step (a job that we do), we need to connect to each input. For each input, this creates a connection between the sites, and a job serving the asset at the remote end. So if we have a step with two inputs both served remotely from different sites, then we have three jobs running at three different sites, and two connections.

From the perspective of the site running the step, we have a local job for the step with a local job id. That is an internal id we use to keep track of our local jobs and give their containers unique names. When we connect to the input assets, we ask the remote site to create a connection for us. It does that, gives it a connection id, and then passes that connection id back to us, so that we can later use it to identify the connection when we want them to shut it down again.

Now, internally, they consider serving an asset to be a job, for which they generate a job id so they can keep the container apart from the ones used for other (step execution or asset serving) jobs. But we don't know that, and anyway we don't consider that job to the the one associated with the connection. For us, that is our local step running job!

And so that's why the two are not the same thing. Now it so happens that the serving site, on shutting down the connection, needs to shut down the data asset container, for which it needs its name, which contains the job id. Job ids are unique within the site, so I can use it as the connection id directly and save myself a translation table from connection ids to job ids or container names. Of course, I do still need to track who requested which connection to make sure people can't shut down each other's connections, but that's a separate thing.

mahiru/components/network_administrator.py
LourensVeen: We'd get in trouble with professor Liskov. She says that since the interface declares that we'll accept a `ConnectionRequest`, an implementing class that only accepts a `WireGuardConnectionRequest` doesn't fully implement the interface. Mypy actually gives a very nice warning:
```
mahiru/components/network_administrator.py:59: error: Argument 3 of "serve_asset" is incompatible with supertype "INetworkAdministrator"; supertype defines the argument type as "ConnectionRequest"
mahiru/components/network_administrator.py:59: note: This violates the Liskov substitution principle
mahiru/components/network_administrator.py:59: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#incompatible-overrides
```

However, I can make the return type more specific (https://en.wikipedia.org/wiki/Covariant_return_type), I'll do that.

mahiru/components/network_administrator.py
LourensVeen: Ah yes, for loading the pilot container. It's really only a few lines though, I'm not sure it's worth a `components/docker.py` with a separate function in it at this point.

mahiru/definitions/connections.py
LourensVeen: It's in the comment for `__init__` I guess. The `net-admin-helper` docs have an explanation as well of how this works; maybe this should be part of the architecture documentation in the future.

LourensVeen: https://github.com/SecConNet/mahiru/issues/103 for that.

PR #98, opened 2021-10-21
Issue 88: 7. Add serving of assets to the AssetStore

LourensVeen:
This adds a reference to the DomainAdministrator to the AssetStore, so that the AssetStore can service requests for connecting to an asset as well as requests for downloading it.

<Usual merge warnings>


svenvanderburg (approved):
Straightforward changes! I like how the `AssetStore` has the job of checking the connection request and raising useful errors, but delegating the serving/stopping logic to the `DomainAdministrator`. This distinction makes sense. No comments!

PR #99, opened 2021-10-21
Issue 88: 8. Extend REST api for asset connections

LourensVeen:
This extends the REST API with new types and endpoints for remotely connecting to assets, and it adds the server-side handlers and a client which implement those.

<Usual merge warnings>

### DIFF
--- a/mahiru/rest/ddm_site.py
+++ b/mahiru/rest/ddm_site.py
@@ -10,8 +10,9 @@ from urllib.parse import quote
 from wsgiref.simple_server import WSGIRequestHandler, WSGIServer
 
 from falcon import (
-        App, HTTP_200, HTTP_201, HTTP_204, HTTP_303, HTTP_400, HTTP_404,
-        Request, Response)
+        App, HTTP_200, HTTP_201, HTTP_204, HTTP_303, HTTP_400, HTTP_403,
+        HTTP_404, Request, Response)
+from jsonschema import ValidationError
 import ruamel.yaml as yaml
 import yatiml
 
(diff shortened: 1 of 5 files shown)


mahiru/rest/ddm_site.py
```diff
@@ -189,6 +190,112 @@ def on_get(
...
+                        request.params['requester'])
+
+                response.status = HTTP_200
+                response.media = serialize(conn_info)
+        except KeyError:
+            logger.info(f'Asset {asset_id} not found')
+            response.status = HTTP_404
+            response.body = 'Asset not found'
+        except RuntimeError:
+            # This is permission denied, but we return a 404 to
```
svenvanderburg, line 236: Nice

mahiru/rest/schemas.yaml
```diff
@@ -422,3 +422,59 @@ components:
...
+          type: integer
+        endpoint:
+          description: WireGuard endpoint for the client side
+          type: object
+          schema:
+            "$ref": "#/components/schemas/WireGuardEndpoint"
+
+    ConnectionRequest:
+      oneOf:
+        - $ref: '#/components/schemas/WireGuardConnectionRequest'
```
svenvanderburg, line 460: Cool, didn't know you could do something like this in `openapi`. I like it!

svenvanderburg (approved):
Looks good!

mahiru/rest/schemas.yaml
LourensVeen: Well, yes. Actually making it work with the Python library requires some acrobatics, and figuring out how to do that requires a dive into the source code because it's not really documented. See https://github.com/p1c2u/openapi-schema-validator/issues/5 :-)

PR #100, opened 2021-10-21
Issue 88: 9. Try to connect to assets before downloading

LourensVeen:
This extends the DomainAdministrator and NetworkAdministrator to try to connect to inputs, then fall back to downloading them and instantiating them locally if that fails.

<Usual merge warning>

### DIFF
--- a/mahiru/components/network_administrator.py
+++ b/mahiru/components/network_administrator.py
@@ -4,13 +4,13 @@ import docker
 import logging
 from pathlib import Path
 from threading import Lock
-from typing import cast, Dict, Tuple
+from typing import cast, Dict, Tuple, Optional
 
 from mahiru.components.settings import NetworkSettings
 from mahiru.definitions.assets import Asset
 from mahiru.definitions.connections import (
-        ConnectionRequest, WireGuardConnectionInfo, WireGuardConnectionRequest,
-        WireGuardEndpoint)
+        ConnectionInfo, ConnectionRequest, WireGuardConnectionInfo,
+        WireGuardConnectionRequest, WireGuardEndpoint)
 from mahiru.definitions.interfaces import INetworkAdministrator
 from mahiru.rest.site_client import SiteRestClient
 
@@ -82,6 +82,7 @@ class WireGuardNA(INetworkAdministrator):
                 request.endpoint)
 
         with self._served_lock:
+            self._served_nets[conn_id] = request.net
             self._served_ports[conn_id] = local_endpoint.port
 
         local_conn_info = WireGuardConnectionInfo(conn_id, local_endpoint)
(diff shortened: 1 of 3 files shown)


mahiru/components/network_administrator.py
```diff
@@ -229,3 +276,12 @@ def _run_net_admin_helper(self, command: str) -> str:
         logs = cast(bytes, container.logs(stdout=True, stderr=False))
         container.remove(force=True)
         return logs.decode('ascii').strip()
+
+    def _net_to_addr(self, net: int, host: int) -> str:
```
svenvanderburg, line 283: Wow :) I would probably use a library for this (because I don't know how IP addresses work), but I think this works and we don't add a dependency.

tests/test_containers.py
```diff
@@ -26,41 +26,36 @@
...
-@pytest.fixture
-def data_asset_tars():
+def get_tar_file(name):
     asset_dir = Path(__file__).parents[1] / 'build'
+    tar_file = asset_dir / name
+    if not tar_file.exists():
+        pytest.skip(
+                f"Image {name} not available. Run 'make assets docker_tars'"
+                " to build the images needed for this test.")
+    return tar_file
```
svenvanderburg, line 36: Nice refactoring!

svenvanderburg (approved):
Looks good! 🥇

mahiru/components/network_administrator.py
LourensVeen: There is no library for it, my concept of `net` (from `net-admin-helper` actually) is entirely custom. The idea is that I don't want to be able to pass arbitrary IP ranges to `net-admin-helper`, only a restricted subset of the `10.x.x.x` private range, because that makes `net-admin-helper` less generally useful, which in turn is a good thing because it's privileged.

The IPv4 `w.x.y.z` notation just writes a 32-bit number as the decimal value of each of its four bytes, separated by dots. The first number is always `10` here because that is a reserved range that we're free to use, and the last bit of `z` is used to distinguish the two endpoints of the connection. That leaves `x`, `y` and the 7 highest bits of `z`. My `net` is those 23 bits expressed as a single integer. A single integer is easy to verify and parse by `net-admin-helper`, so there's less chance of a buffer overflow or something there. Different net numbers mean different, non-overlapping IP addresses for the connection, so if the compute container picks a different net number for each connection then we can be sure that there are no collisions and each input can be reached on a unique IP.

So Mahiru does just that, and deals only in net numbers. The only place where the IP address is still needed is inside of the compute asset, because it needs to connect to the remote container on an IP address, and it doesn't know anything about my net numbers. So that's where we have to calculate the corresponding IP address, and that's where this function comes in.

These nets are actually documented at https://github.com/SecConNet/net-admin-helper/blob/develop/docs/container_wireguard.md, in somewhat more technical language. Maybe I should point to that in the comment.

LourensVeen: And actually, why did I think Python didn't have bitwise operators? It does, and I rewrote this to use them, because it shows much more clearly what's going on.

LourensVeen: And actually, why did I think Python didn't have bitwise operators? It does, and I rewrote this to use them, because it shows much more clearly what's going on.

LourensVeen:
And then when all is said and done and you're ready to merge you discover a bug, and then the fix doesn't end up in the new version because you forgot to `git add` and `git commit --amend` will happily pretend to do something even though there aren't any staged changes, and then you do the whole thing again and then you can finally merge. Urgh. But oh well, all's well that ends well :-).

PR #101, opened 2021-10-21
Issue 88: 10. Add a remote data access scenario

LourensVeen:
This adds a new scenario to the test/demo scenarios which demonstrates remote asset access. It actually runs the exact same workflow as the compute-to-data scenario, but the policies are set differently: the compute is not allowed to move, and the data is. Thus, the planner will set up a data-to-compute scenario, and the system will try to connect to the asset remotely. This is enabled in this case, so it succeeds.

This has a fair amount of copy-paste from the compute-to-data version, but I think it's nice if people can look at one scenario at at time without having to jump around to shared code in other directories. I'd like your opinion on that though.

#include <merge_warning.h>

### DIFF
--- a/scenarios/data_access/client1/init_site.py
+++ b/scenarios/data_access/client1/init_site.py
@@ -1,0 +1,115 @@
+#!/usr/bin/env python3
+
+from pathlib import Path
+
+from cryptography.hazmat.backends import default_backend
+from cryptography.hazmat.primitives.asymmetric.rsa import (
+        generate_private_key, RSAPrivateKey, RSAPublicKey)
+from cryptography.hazmat.primitives.serialization import (
+    Encoding, PublicFormat)
+import requests
+
+from mahiru.definitions.assets import ComputeAsset, ComputeMetadata, DataAsset
+from mahiru.definitions.registry import PartyDescription, SiteDescription
+from mahiru.policy.rules import (
+        MayAccess, ResultOfComputeIn, ResultOfDataIn)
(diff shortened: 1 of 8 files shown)


scenarios/data_access/client1/init_site.py
```diff
@@ -0,0 +1,115 @@
...
+            f'{ASSET_DIR}/compute-asset.tar.gz',
+            script_metadata)
+
+    assets = [output_base, script]
+
+    for asset in assets:
+        client.store_asset(asset)
+
+
+def add_initial_rules(
```
svenvanderburg, line 69: Maybe you can add some documentation here. Maybe in the form of small natural language comments like "site1 may access it's own output asset". Because these rules basically define the behaviour of the system right? Also might be nice to add a short description of the bigger picture here, I quite like the one in the description of this PR.

svenvanderburg (approved):
Yeah, the story is finished! Great work @LourensVeen! 🥳 Nice to see all the actors in the novel coming together here (although they are actually behind the screen in this scenario).

You could consider adding a bit more documentation. Maybe a Readme to the scenarios/data_access describing the scenario, or the suggestion in my comment. All up to you of course!

Indeed I think the scenario code should be sort of stand-a-lone, so people can follow it in itself.

LourensVeen:
It's a good point, the other one is also not documented. There's generic documentation on how to run them, but no specific description of each scenario. I've made https://github.com/SecConNet/mahiru/issues/104 for this.

Issue #109, opened 2022-02-17
Namespaces should be owned by parties, not by sites


LourensVeen:
We currently assume that each namespace has a unique site that serves its policies, and we associate that namespace with that site. But that's not really right, namespaces should be owned by a party (and in the party's certificate, when we get those). So a better system would be to have the namespaces be owned by parties, and to have sites have an optional namespace for which they serve policies. Then if a party has two sites, both could serve some of the policies for the same namespace, or the party could register two sub-parties with separate namespaces. 

This does pose an issue with the policy replication system, which currently assumes that each namespace has a unique site that serves its policies. But we can organise the replicas per site rather than per namespace, that should work.

- [x] Update test cases
- [x] Update `PartyDescription` and serialisation to add namespace
- [x] Update `SiteConfiguration` comments to indicate that this is the namespace the site is in
- [x] Make `PolicyClient` organise policy replicas by site id rather than namespace
- [x] Make `RegistryClient.get_public_key_for_ns` look up the party directly and get its namespace
- [x] Make `Registry.register_site` check that the site's namespace, if any, matches that of its owner

LourensVeen:
Actually, why would a site have a namespace at all? Why not just make it so a site can serve only its owner's namespace? We could still make serving policies optional, but we don't want sites to serve something other than their owner's namespace, and we're not doing sub-namespaces. So why not remove the namespace from the `SiteDescription` while we're at it, and replace it with a boolean `serves_policies`?

- [x] Replace `namespace` with `serves_policies` in `SiteDescription` and serialisation
- [x] Update `PolicyClient` to use it
- [x] Update tests

Oh, and by the way, we have to update the example scenarios to match these changes:
- [x] Update scenarios

PR #114, opened 2022-02-28
Issue 109: Namespaces owned by parties - part 1

LourensVeen:
This implements the first part of #109.

I'm experimenting a bit with the set-up of this PR. The idea is to build the feature in a few easy-to-review stages, with each stage's commits on let's-call-it-a staging branch that gets fast-forward merged into the feature branch upon a positive review. Once all staging branches have been merged, the feature branch is completed and completely reviewed, so it can be non-fast-forward merged into develop without another review. Easy to review and a clean git history!

Except of course if I discover during the final stage that I should have done something differently in the beginning, in which case we have to either rebase and re-review, or put a patch on top. We'll see how it goes :-).

### DIFF
--- a/mahiru/components/policy_client.py
+++ b/mahiru/components/policy_client.py
@@ -2,6 +2,7 @@
 from typing import Dict, Iterable, Set
 
 from mahiru.components.registry_client import RegistryClient
+from mahiru.definitions.identifier import Identifier
 from mahiru.definitions.interfaces import IPolicyCollection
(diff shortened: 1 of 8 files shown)


LourensVeen:
Oops, forgot to update the REST API definitions, those are only tested by the scenario tests/demos so I didn't notice.

svenvanderburg (approved):
Looks good 👍  I also like the small naming improvements (i.e. `create_keys` vs `create_parties`).

Let's have a party once this issue is done...

PR #115, opened 2022-03-01
Issue 109: Namespaces owned by parties - part 2

LourensVeen:
Here's the second part of #109 , now removing the namespaces from the sites. Instead, sites may either serve policies or not, and if they do those policies have to be within the owning party's namespace. This also fixes the scenarios to work with these changes and the ones in part 1.

### DIFF
--- a/mahiru/rest/schemas.yaml
+++ b/mahiru/rest/schemas.yaml
@@ -362,9 +362,9 @@ components:
         - owner_id
         - admin_id
         - endpoint
-        - runner
-        - store
-        - policies
+        - has_runner
+        - has_store
+        - has_policies
       properties:
         id:
           description: Identifier of the site
@@ -380,13 +380,13 @@ components:
             REST endpoint at which the site may be contacted.
           type: string
           format: uri
-        runner:
+        has_runner:
           description: Whether the site has a runner
           type: boolean
-        store:
+        has_store:
           description: Whether the site has a store
           type: boolean
-        policies:
+        has_policies:
           description: Whether the site serves policies
           type: boolean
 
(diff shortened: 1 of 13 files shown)


mahiru/rest/schemas.yaml
```diff
@@ -386,12 +386,9 @@ components:
...
           type: boolean
-        namespace:
-          description: >-
-            Namespace for which the policy server serves policy, if there
-            is a policy server at this site, otherwise null.
-          type: string
-          nullable: true
+        policies:
+          description: Whether the site serves policies
+          type: boolean
```
svenvanderburg: I was confused for a small moment because `policies` somehow implicates a list of `policy` elements. But the type field makes it clear this is a boolean. but maybe it is an idea to change `policies` to `serves_policies` or something like that.

tox.ini
```diff
@@ -19,7 +19,7 @@ commands =
     pydocstyle mahiru
 
 [testenv:docs]
-descriptionn = Build documentation using Sphinx
+description = Build documentation using Sphinx
```
svenvanderburg, line 22: 🎉

svenvanderburg (approved):
Looks good. I just have one comment about naming.

mahiru/rest/schemas.yaml
LourensVeen: Good point. I did it like this for symmetry with `runner` and `store`, which should probably become `has_runner` and `has_store` then. I'll fix it, why not :-).

tox.ini
LourensVeen: Every little bit counts!

Issue #87, opened 2021-07-02
Improve policies


LourensVeen:
This probably needs some sub-issues, and then this will collect them all in one place here so we can track overall progress a bit.

Output names:
- [x] Add output name to ResultOfDataIn
- [x] Add output name to ResultOfComputeIn

Categories:
- [x] Add asset categories (InAssetCategory rule)
- [x] Add category traversal for compute asset in ResultOfDataIn rules (instead of collection, if already there?)
- [x] Add category traversal for data asset in ResultOfComputeIn rules, instead of categories
- [x] Add collection traversal for compute asset in ResultOfComputeIn (in not already there?)
- [x] Add site categories (InSiteCategory rule)
- [x] Add category traversal for sites in MayAccess rules
- [x] Add party categories (InPartyCategory rule)

More control:
- [x] Add MayUse rule

LourensVeen:
Hurray, all done!

PR #116, opened 2022-03-07
Issue 87: Improve policies - Part 1: outputs

LourensVeen:
This is the first part of the implementation of policies V2. It adds an `output` attribute to `ResultOfIn` rules, and modifies the evaluation rules so that the rules apply to the specified output. Handy for steps that produce some sensitive and some non-sensitive outputs.

Partially implements #87.

### DIFF
--- a/mahiru/policy/rules.py
+++ b/mahiru/policy/rules.py
@@ -122,6 +122,7 @@ class ResultOfIn(Rule):
             self,
             data_asset: Union[str, Identifier],
             compute_asset: Union[str, Identifier],
+            output: str,
             collection: Union[str, Identifier]
             ) -> None:
         """Create a ResultOfIn rule.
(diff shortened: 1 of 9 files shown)


mahiru/policy/rules.py
```diff
@@ -122,13 +122,15 @@ def __init__(
...
             compute_asset: Union[str, Identifier],
+            output: str,
             collection: Union[str, Identifier]
             ) -> None:
         """Create a ResultOfIn rule.
 
         Args:
             data_asset: The source data asset.
             compute_asset: The compute asset used to process the data.
+            output: The workflow output that produced the result Asset.
```
svenvanderburg: So what you mean with 'output' refers to sort of the variable name of the workflow output value that is then feed into the result Asset? It's a bit implicit from the docstring alone, this is what I figured from the examples. Maybe you can add a bit more details here or add it to your 'glossary' overview.

svenvanderburg (approved):
Looks good. I just added one comment for clarification.

mahiru/policy/rules.py
LourensVeen: Yes, it is the name of the output of the workflow step that produced the asset that's going to be in the collection. I've tried to clarify that a bit, how's this?

svenvanderburg: Yes, looks good! :)

PR #117, opened 2022-03-10
Issue 87: Improve policies - Part 2: Asset categories

LourensVeen:
This adds the `InAssetCategory` rule for putting assets into categories, modifies the policy evaluation algorithm to use them where appropriate, and adds a bunch of unit tests for policies.

The unit tests are a bit repetitive, I've tried to simplify them but it doesn't get much better. But it's a bunch of tests, some repetition is okay.

### DIFF
--- a/mahiru/definitions/identifier.py
+++ b/mahiru/definitions/identifier.py
@@ -13,7 +13,8 @@ class Identifier(str):
     3. site:<namespace>:<name>
     4. asset:<namespace>:<name>:<site_namespace>:<site_name>
     5. asset_collection:<namespace>:<name>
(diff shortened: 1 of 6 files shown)


svenvanderburg (approved):
Looks good Lourens! I like the idea of asset categories :) Indeed the unittests are a bit repetitive, but I think this is the best way to go about it. You could split out common things in functions or test multiple related things in the same test function, but this is much clearer: every test just tests slightly different scenario and you can easily see how the scenario is build up.

LourensVeen:
Ugh, messed up the reference branch again...it's been merged alright.

PR #120, opened 2022-04-08
Issue 87: Improve policies - Part 3b: Refactor

LourensVeen:
This does a bit of refactoring. The PartyCollections we had turned out to make no sense, so that's been turned into PartyCategory, which we do need. Then there's a bug fix in the access checking, which didn't walk the hierarchy on one side as a result of which certain things that should have been allowed were forbidden. Finally, the functions for walking the hierarchy had a lot of overlap, and I saw a fairly clean way of combining them, so I did. Part of #87.

### DIFF
--- a/mahiru/policy/evaluation.py
+++ b/mahiru/policy/evaluation.py
@@ -1,6 +1,7 @@
 """Components for evaluating workflow permissions."""
 from typing import (
-        Dict, Iterable, List, Optional, Set, Union, Tuple, Type, TypeVar)
+        Callable, Dict, Iterable, List, Optional, Set, Union, Tuple, Type,
+        TypeVar)
 
 from mahiru.definitions.identifier import Identifier
 from mahiru.definitions.interfaces import IPolicyCollection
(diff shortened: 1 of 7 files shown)


mahiru/policy/evaluation.py
```diff
@@ -109,79 +119,42 @@ def matches_one(
...
-            for party in cur_parties:
-                for rule in self._policy_collection.policies():
-                    if isinstance(rule, InPartyCollection):
-                        if rule.party == party:
-                            new_parties.append(rule.collection)
-        return cur_parties
-
-    def _upward_equivalent_objects(
-            self, rule_type: Type[_GroupingRule], obj: Identifier
+    def _equivalent_objects(
```
svenvanderburg, line 126: Nice refactoring!

mahiru/policy/rules.py
```diff
@@ -109,39 +109,47 @@ def signing_namespace(self) -> str:
         return self.category.namespace()
 
 
-class InPartyCollection(Rule):
-    """Says that Party party is in PartyCollection collection."""
+class InPartyCategory(GroupingRule):
+    """Says that a PartyCollection contains a Party."""
```
svenvanderburg:
```suggestion
    """Says that a PartyCategory contains a Party."""
```
right?

svenvanderburg (approved):
See my small comments, but mostly: bueno!

PR #121, opened 2022-04-08
Issue 87: Improve policies - Part 4: may use

LourensVeen:
This finally adds the usage checking functionality, the last bit of #87. To check that the submitter of the workflow has use rights to the outputs, we need to know who it is, so we add the submitting party to the Job class. Sending an identifier is obviously not very secure, but this will be replaced by a signature when we add certificates. Then there's the MayUse rule, and the logic for evaluating it.

### DIFF
--- a/mahiru/policy/evaluation.py
+++ b/mahiru/policy/evaluation.py
@@ -7,8 +7,8 @@ from mahiru.definitions.identifier import Identifier
 from mahiru.definitions.interfaces import IPolicyCollection
 from mahiru.definitions.workflows import Job, Plan, Workflow, WorkflowStep
 from mahiru.policy.rules import (
-        GroupingRule, InAssetCategory, InAssetCollection,
-        InSiteCategory, MayAccess, ResultOfIn, ResultOfDataIn,
+        GroupingRule, InAssetCategory, InAssetCollection, InPartyCategory,
+        InSiteCategory, MayAccess, MayUse, ResultOfIn, ResultOfDataIn,
         ResultOfComputeIn)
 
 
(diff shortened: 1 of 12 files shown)


mahiru/policy/rules.py
```diff
@@ -227,6 +227,42 @@ def signing_namespace(self) -> str:
...
+    """Says that Party party may use Asset asset."""
+    def __init__(
+            self, party: Union[str, Identifier], asset: Union[str, Identifier],
+            conditions: str) -> None:
+        """Create a MayUse rule.
+
+        Args:
+            party: The party that may access.
+            asset: The asset that may be accessed.
+            conditions: Conditions under which the asset may be used.
```
svenvanderburg, line 240: `conditions` is not enforced by the program right? Maybe document that this is just a description of the conditions, but that it is not checked.

mahiru/policy/evaluation.py
```diff
@@ -123,6 +123,40 @@ def matches_one(
         return all([matches_one(asset_set, equiv_sites)
                     for asset_set in permissions._sets])
 
+    def may_use(self, permissions: Permissions, party: Identifier) -> bool:
+        """Checks whether an asset can be used by a party.
+
+        This function checks whether the given party has use rights
+        to at least one asset in each of the given set of assets.
```
svenvanderburg, line 130: I find it a bit non-inutitive that a party may use all assets if there is just one asset in the set that the party may use (even this sentence is contradicting in itself). But since you are Lourens I assume that there is a good reason for it :)

svenvanderburg (approved):
Looks good! 👍  Just a few comments for clarification.

mahiru/policy/rules.py
LourensVeen: Yes, that's correct, it's a friendly request, or a reminder of a separately agreed upon contract, or an open source copyright license, but this applies when the asset leaves the system, so Mahiru cannot enforce it.

I'm not sure that this comment is the best place to explain this though, I think this belongs in the documentation and in some kind of hint in a future GUI.

mahiru/policy/evaluation.py
LourensVeen: Haha, well thanks for the vote of confidence :-).

If you give someone access to one of your collections, then you give them access to all the assets in it and its subcollections. That means that to have access to a specific asset, it's enough to have access to that asset directly, or to any of the collections it is directly or indirectly in. It is this set of what I'm calling equivalent assets that is passed in here, so it is indeed enough to have access to only one of them.

The result of a processing step is considered to be a result of each of its inputs as well as of the software used. So for each input and the software, the owner needs to say in which collection they deem the result to be in, and then give access to that collection (or a containing collection). So we actually get a set like the one we have here for each input and the software, and you need to have access to (at least one asset in) _each_ of those to be able to access the output.

(We computer scientists call this a _conjunctive normal form_ when we're trying to impress a date. It doesn't work.)

PR #122, opened 2022-05-31
Issue 119: Include rule types in signature hashes

### LINKED ISSUE #119: Rule signature hashes should include rule type
Currently, the signing representations for e.g. ResultOfDataIn and ResultOfComputeIn are identical. If a party is owner of both the data asset and the compute asset, and sets a ResultOfComputeIn but no ResultOfDataIn for a given output, then someone else could inject that ResultOfDataIn and copy the signature, and it would validate. And that extra rule may enable a workflow that would otherwise be forbidden. So we need to include the rule type in the signature for all rules.

LourensVeen:
This fixes #119, and while we're here adds identifier type checks on the rule classes so that when we receive rules from others, we'll check to make sure that they're not sending us nonsense.

### DIFF
--- a/mahiru/components/settings.py
+++ b/mahiru/components/settings.py
@@ -81,6 +81,10 @@ class SiteConfiguration:
             loglevel: Logging level to use, one of 'critical', 'error',
                     'warning', 'info', or 'debug'.
         """
+        if owner.kind() != 'party':
+            raise ValueError(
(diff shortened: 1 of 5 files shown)


svenvanderburg (approved):
Nice work, good that this is fixed! (Of course I should have spotted this before... 🙄 ). I have no futher comments.

